PR #29, opened 2022-02-24
allow external regressors

jmoralez:
Allows for extra columns in the dataframe to be used as external regressors when training and adds an argument `xreg` to the `forecast` method to pass the future values of the regressors.


jmoralez:
We should probably make sure that xreg is sorted by id and date (which is assumed but not checked yet). WDYT @FedericoGarza?

FedericoGarza (approved):

FedericoGarza:
> We should probably make sure that xreg is sorted by id and date (which is assumed but not checked yet). WDYT @FedericoGarza?

Yes, I think it would be great to have that check before computing forecasts. :)

mergenthaler:
@all-contributors please add @jmoralez for code

PR #33, opened 2022-02-24
Feat: xreg for auto_arima

FedericoGarza:
- fix: add xreg functionality for arima
- fix: xreg substraction (was a hard one xD)
- allow external regressors
- add default to auto_arima future_xreg
- extract first column only if input is 2d
- use boolean mask instead of groupby.head
- feat: add forecast_arima fn for Arima fn
- fix: add mising bestfit keys auto_arima_f
- fix: auto_arima_f bugs for xreg
- feat: add external regressors auto_arima (and index example)


FedericoGarza:
I had to use `.values` or otherwise the function receives a `DataFrame`; some models can handle it, but especially `auto_arima` receives an array as future_xreg. What do you think @jmoralez? 

https://github.com/Nixtla/statsforecast/blob/c1003e5f4ec78398e2a225aacb6db6e001604ebe/statsforecast/core.py#L55-L57

jmoralez:
Yeah I think we should do some preprocessing on xreg in the forecast method, maybe we can have xreg have the same structure as the input df but without the y column, sort it by id and date and then turn it into an array before passing it to the compute_forecast method. We could also validate the shape there (nrows should be n_series * h and ncols should be self.ga.data.shape[1])

jmoralez:
Can you please also clean the notebooks? Saving the outputs can increase the size of the repo very quickly.

FedericoGarza:
Hi @jmoralez! I added preprocessing for `xreg` allowing the `ds` column. The preprocessing is the same as done in `_grouped_array_from_df`. I also cleaned the nbs. :)

README.md
```diff
@@ -203,6 +203,103 @@ for label in (ax.get_xticklabels() + ax.get_yticklabels()):
...
+        'ds': pd.date_range(start='1960-01-01', periods=ap_test.size, freq='M')
+    },
+    index=pd.Index([0] * ap_test.size, name='unique_id')
+)
+```
+
+```python
+xreg_test['trend'] = np.arange(133, ap_test.size + 133)
+xreg_test['month'] = xreg_test['ds'].dt.month
+xreg_test = pd.get_dummies(xreg_test, columns=['month'], drop_first=True)
```
jmoralez, line 250: I think we should add an intercept or leave the first dummy here.

jmoralez (changes requested):
LGTM! I just left some minor comment on the example using xreg.

README.md
FedericoGarza: Cool, solved here 5dd8ebb8282a515cce324cf9eeb7a3999b09463c.

jmoralez (approved):

PR #35, opened 2022-02-28
fixes/arima

FedericoGarza:
- fix: xreg usage in arima_css_op
- fix: polyroot conditions inside myarima
- fix: add constant to OLS residuals auto_arima_f
- fix: add arima module fixes


mergenthaler:
@all-contributors please add @FedericoGarza for code

statsforecast/arima.py
```diff
@@ -1193,11 +1202,11 @@ def myarima(
             else:
                 last_nonzero = 0
             if last_nonzero > 0:
-                testvec = testvec[:last_nonzero]
-                proots = np.polynomial.polynomial.polyroots(np.append(1, -testvec))
```
jmoralez, line 1197: This is originally
```R
testvec <- testvec[1:last.nonzero]
proots <- try(polyroot(c(1, -testvec)))
```
So it's the first last_nonzero elements and then get the roots of the negative with a -1 at the front. Why was this wrong?

jmoralez: Ah I see, I didn't see this was for the AR part. I'm not sure about the limits yet, in R vec[1:n] returns the first n elements, and in Python vec[:n] also does. So I think it should be testvec[:last_nonzero], because testvec[:(last_nonzero + 1)] would be equal to testvec[1:(last.nonzero+1)] in R

FedericoGarza: I think the problem is with zero indexing, in the following example python returns `last_nonzero=4`, so in reality returns `n-1` and the slicing returns `testvec[:(n-1)]` instead of `testvec[:n]`

image

jmoralez (approved):

PR #54, opened 2022-03-04
add confidence intervals for arima

jmoralez:


FedericoGarza (approved):
🎉 LGTM. Let's add an example to `index.ipynb` plotting the intervals as is the [fable](https://fable.tidyverts.org/articles/fable.html) or forecast package because right now it is hard to read the results. :)

PR #65, opened 2022-03-08
Fix parallel processing bugs

FedericoGarza:
- fix: parallel processing
- fix: n_jobs greater than n_series
- fix: build library changes


jmoralez:
The ProcessPoolExecutor needs the `if __name__ == '__main__'` safeguard. I haven't tried it but I guess we can add that to the cells that test with `n_jobs > 1` so that it doesn't crash.

jmoralez (approved):

PR #73, opened 2022-03-10
feat: add AutoARIMA class

FedericoGarza:
I'm following the `pmdarima` API.


statsforecast/arima.py
```diff
@@ -2085,4 +2086,284 @@ def try_params(p, d, q, P, D, Q, constant, k, bestfit):
...
+
+    def predict(self, h: int, X: Optional[np.ndarray] = None,
+                level: Optional[int] = None):
+        """Forecast future values using a fitted AutoArima.
+
+        Parameters
+        ----------
+        h: int
+            Number of periods for forecasting.
+        X: array-like of shape (n, n_x) optional (default=None)
```
jmoralez, line 2347:
```suggestion
        X: array-like of shape (h, n_x) optional (default=None)
```

statsforecast/arima.py
```diff
@@ -2085,4 +2086,284 @@ def try_params(p, d, q, P, D, Q, constant, k, bestfit):
...
+            Future exogenous variables.
+        level: int
+            Confidence level for prediction intervals.
```
jmoralez, line 2350: Maybe this can be an optional tuple instead to get several levels at once

jmoralez (commented):
Nice! This will be very useful to build stuff into, like the summary and other requested features. I left some comments and I think it'd be nice to use True and False in the docs as well as specifying the defaults for the parameters that aren't optional.

statsforecast/arima.py
FedericoGarza: Nice! I think the second option is better, I'll include it :)

jmoralez:
What do you think about returning a pandas DataFrame or a dict from the predict method? I think currently the intervals are sorted increasingly (lo 95, lo 80, hi 80, hi 95) which might be a bit unexpected. Maybe having the named outputs is easier on the user.

FedericoGarza:
Cool, always a dataframe sounds good

FedericoGarza:
Done @jmoralez :)

jmoralez (approved):

PR #86, opened 2022-04-11
Add autoarima summary and repr

FedericoGarza:
- feat: add arima summary class
- feat: add autoarima summary


nbs/arima.ipynb
```diff
@@ -3018,6 +3018,69 @@
     "mod = auto_arima_f(ap, period=12, method='CSS-ML', trace=True)"
```
mergenthaler, line 3018: <p>Maybe include something in cases those conditionals are not satisfied. Something like "This model does not have lamdas, coefs or loglik, be sure its properly trained."</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/statsforecast/pull/86/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='83'/>

FedericoGarza: Included in 279def9

PR #106, opened 2022-05-04
Add dependency hint for quick intro

guerda:
The "how to use" section uses ipython and matplotlib which are not automatically installed with statsforecast. Therefore, it's recommendable to install both packages on top of it.


FedericoGarza:
Thank you @guerda! Welcome to `StatsForecast`🎉

PR #111, opened 2022-05-08
Add fugue example

goodwanghan:


goodwanghan:
@FedericoGarza @mergenthaler I just added a performance experiment using Fugue + Spark/Dask because I feel this is very interesting. In the notebook, I only publicized the performance numbers on 10k serieses and 128 cpus. If you are interested we can chat more about the performance on larger datasets. I think it would be fair if you conduct the experiments.

FedericoGarza:
Thank you @goodwanghan! And sorry for the late response. I also opened an [issue](https://github.com/Nixtla/statsforecast/issues/119) to include fugue as a native backend for StatsForecast. 

Welcome to StatsForecast :tada: :D

FedericoGarza:
Once we have the first version, we will conduct the experiments. :)

goodwanghan:
@FedericoGarza that sounds great, thank you!

PR #143, opened 2022-07-15
fix: lik float

AzulGarza:


FedericoGarza:
The problem arises with Python3.6. We will deprecate it.

PR #172, opened 2022-08-02
Fix column name issues with seasonalities

goodwanghan:


FedericoGarza:
Hey @goodwanghan! 

As we have changed the model's name as of v1.0.0, this fix is no longer necessary. Thank you!

Issue #159, opened 2022-07-25
boxcox not imported


EBM-JH:
**Describe the bug**
For AutoARIMA and presumably ARIMA too, when you set a blambda parameter, the code produces the below error:

NameError: name 'boxcox' is not defined

It appears that this is due to scipy.stats.boxcox not being imported.

My proposed fix would be to change line 19 from:

from scipy.stats import norm

to

from scipy.stats import norm, boxcox

I haven't tested it, sorry. Could be the simplest bug fix ever!

**To Reproduce**
1. Try and fit a model with blambda=0

**Expected behavior**
NameError: name 'boxcox' is not defined

**Desktop (please complete the following information):**
 - OS: Win 10
 - Browser Chrome
 - Version 103

**Additional context**
I came across this running darts.models.forecasting.sf_auto_arima.StatsForecastAutoARIMA, which is a wrapper for this package.

PR #174, opened 2022-08-02
Import boxcox in arima.py

luthfianto:
Related to #159, I have the same problem too


luthfianto:
@Nixtla what should we do with the ` setattr(blambda, 'biasadj', biasadj)` line?

Issue #32, opened 2022-02-24
[question] Division by Zero error 


tackes:
When running on many groups of time series, some groups are giving me a 'division by zero' error, and the script stops.
Is there a way to pass through this error and complete the forecasts without errors?

FedericoGarza:
Hi @tackes! Would it be possible for you to share with us any time series that gives you the error? To try to replicate and debug it :)

tackes:
The issue is with short time series.  The below will return 2 errors - 
with 
1) random_walk_with_drift will return a **divide by zero** error.
2) arima will return **math domain** error

In the code sample, one removing the 'test' period, it will only have 1 data point to do calculations. 
What would be a nice feature is if there was an option to 'ignore errors' and keep processing.  This particular time series was on the 812th time series group out of 2000, but because it hit an error, the entire script stops running.
Even better is if there was a logging feature to store what and where the error occurred.


`data = {'unique_id': ['A','A','A','A','A'],
        'y':[128050,4200,0,0,0],
        'ds': pd.date_range(end='2021-09-01', periods=5, freq='M')}
series = pd.DataFrame(data,
            columns =['unique_id','y','ds'])

series_test = test_df.groupby('unique_id').tail(horizon).copy()
series = test_df.drop(series_test.index)
series['unique_id'] = series['unique_id'].astype('object')

series = series.set_index('unique_id')
seasonality = 12 #Monthly data
models = [   random_walk_with_drift,   auto_arima]

fcst = StatsForecast(df=series, models=models, freq='M', n_jobs=1)
forecasts = fcst.forecast(horizon)
forecasts.reset_index()`

mergenthaler:
@all-contributors please add @tackes for bug

blazespinnaker:
Get this error as well quite frequently, 

StatsForecast(
            models = [AutoARIMA(season_length = 3)],
            freq = 'M')


 unique_id ds y
40041     2021-10-01  2.607948
40041     2021-09-01  2.613875
40041     2021-08-01  2.646475
40041     2021-07-01  2.667220
40041     2021-06-01  2.673147
40041     2021-05-01  2.705746
40041     2021-04-01  2.693892
40041     2021-03-01  2.735382
40041     2021-02-01  2.690929
40041     2021-01-01  2.578313

PR #18, opened 2022-02-19
Readme | update yml | included CONTRIBUTING

mergenthaler:

PR #201, opened 2022-08-18
Intervals for some simple methods

MMenchero:


nbs/models.ipynb
```diff
@@ -57,6 +57,7 @@
     "import pandas as pd\n",
```
FedericoGarza, line 57: <p>Since <code>quantiles</code> is a private function (only used inside classes trough <code>_calculate_intervals</code>), can we change <code>quantiles</code> to <code>_quantiles</code>?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/statsforecast/pull/201/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

FedericoGarza, line 57: <pre class="ql-syntax" spellcheck="false">Line #7.    def _calculate_intervals(out, level, h, sigmah, fitted):
</pre><p>The argument <code>fitted</code> is not used in the function, so you can remove it.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/statsforecast/pull/201/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

FedericoGarza, line 57: <pre class="ql-syntax" spellcheck="false">Line #8.            return f'HistoricAverage()'
</pre><p>Please remove the parenthesis after <code>HistoricAverage</code> (it was a previous change now fixed in main). </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/statsforecast/pull/201/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='89'/>

FedericoGarza, line 57: <pre class="ql-syntax" spellcheck="false">Line #18.            sigma = np.sqrt(1/(len(y)-1)*np.nansum(np.power(mod['residuals'],2)))
</pre><p>This line is a little bit hard to read. Maybe we can divide it, something like:</p><p> </p><pre class="ql-syntax" spellcheck="false">sigma = np.nansum(mod['residuals'] ** 2)
sigma = sigma / (len(y) - 1)
sigma = np.sqrt(sigma)
</pre><p> </p><p>Since these lines are repeated for each model, maybe we can create a new function, before <code>_quantiles</code>, to prevent errors:</p><p> </p><pre class="ql-syntax" spellcheck="false">def _calculate_sigma(residuals: np.ndarray):
    n = len(residuals) #residuals and y have the same length
    sigma = np.nansum(residuals ** 2)
    sigma = sigma / (n - 1)
    sigma = np.sqrt(sigma)
    return sigma
</pre><p> </p><p>And use it whenever necessary.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/statsforecast/pull/201/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='89'/>

FedericoGarza, line 57: <pre class="ql-syntax" spellcheck="false">Line #32.            self.model_['mean'] = mean
</pre><p>Let's remove this line, to avoid updating the object <code>self.model_</code> . Let's use <code>res</code> in <code>_calculate_intervals</code> instead.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/statsforecast/pull/201/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='89'/>

FedericoGarza, line 57: <pre class="ql-syntax" spellcheck="false">Line #56.            out = _historic_average(y=y, h=h, fitted=True)
</pre><p>Instead of <code>fitted=True</code>, I think we can use:</p><p> </p><pre class="ql-syntax" spellcheck="false">fitted = fitted or (level is not None)
out = _historic_average(y=y, h=h, fitted=fitted)
</pre><p>   </p><p>Then, if the fitted values are necessary for the calculation of the prediction intervals, <code>_historic_average</code> will return them even if <code>fitted=False</code>. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/statsforecast/pull/201/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='89'/>

FedericoGarza, line 57: <p>It would be great to add a unitary test, something like:</p><p> </p><pre class="ql-syntax" spellcheck="false">fcst_ha = ha.forecast(ap, 12, None, None, False, (80,95))
test_eq(
    fcst_ha['lo-80'],
    #here goes the expected value for lo-80
)
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/statsforecast/pull/201/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='90'/>

FedericoGarza (changes requested):
Nice job, @MMenchero! I left some comments. Most of them are related to `historic_average`, but they also apply to the rest of the models.

nbs/models.ipynb
kdgutier: :)

MMenchero: <p>This was implemented and used for the historic average, the naive, and the rwd. However, the seasonal naive uses </p><p> </p><p><code>sigma = sigma/(n-season_length)</code></p><p> </p><p>So in this case I computed sigma directly in the function. </p>

MMenchero: <p>Done. Also used res in the other models. </p>

MMenchero: <p>Used <code>np.testing.assert_almost_equal(... decimal=4)</code> instead since Hyndman's functions round results up to 4 decimals and Python doesn't, </p>

MMenchero: <p>The <code>_calculate_sigma</code> function will be changed so that it can be used in all 4 simple methods. </p>

FedericoGarza (approved):

PR #294, opened 2022-11-02
[FIX] Sparse models with zero-valued time series

### LINKED ISSUE #258: [BUG] Croston Fails to fit intermittent time series
**Describe the bug**
A user reported [this bug](https://stackoverflow.com/questions/73946620/error-in-data-frame-definition-while-multiple-ts-stat-forecasting-in-python). The problem arises due to the Croston models. Omitting them solves the issue.

**To Reproduce**
Install `statsforecast==0.6.0`.

**Additional context**
The version before `1.0.0` is no longer maintained. We must try reproducing the code using `statsforecast>=1.0.0` and see if the problem persists.

FedericoGarza:
FIX #258


nbs/models.ipynb
```diff
@@ -4057,6 +4057,16 @@
     "        return out"
```
FedericoGarza, line 4057: <pre class="ql-syntax" spellcheck="false">Line #15.        if yip == 0.:
</pre><p>Please fix this. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/statsforecast/pull/294/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='195'/>

mergenthaler (approved):

nbs/models.ipynb
FedericoGarza: fixed in [07f9aba](https://github.com/Nixtla/statsforecast/pull/294/commits/07f9aba0f795c94e90b13a9c44edaaae7f07eb26)

PR #301, opened 2022-11-10
[FIX] Add explicit optional argument (PEP-484)

FedericoGarza:
This PR fixes the lint error:

```
PEP 484 prohibits implicit Optional. Accordingly, mypy has changed its default to no_implicit_optional=True
```﻿


MMenchero (approved):
Everything looks ok to me.

PR #302, opened 2022-11-10
[FIX] SeasonalNaive

FedericoGarza:
This PR fixes the SeasonalNaive model.

The previous code assumed that the series observations were multiples of the `season_length`. ﻿

I've added some tests.


MMenchero (approved):
These changes should fix the issue with the `Seasonal Naive`.

PR #304, opened 2022-11-14
[FEAT] add precompile option to statsforecast class

FedericoGarza:
This PR adds the `precompile` argument to `StatsForecast` to compile models with a shorter time series.﻿


nbs/core.ipynb
```diff
@@ -65,7 +65,9 @@
     "import matplotlib.pyplot as plt\n",
```
mergenthaler, line 65: <pre class="ql-syntax" spellcheck="false">Line #14.                precompile: bool = True
</pre><p>I think adding this as an parameter of the class makes it more complex for the end user. It's not clear for users when to precompile. </p><p> </p><p>The idea behind this change is to precomplie the Models with the smallest dataset possible. Currently your test cover most use cases and you might be "overcontrolling" for edge cases. I think it would be great if we could leave this part inside the class and try to catch possible errors.</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/statsforecast/pull/304/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='28'/>

PR #313, opened 2022-11-18
[FEAT] Add plotly as engine to StatsForecast's plot method

FedericoGarza:
This PR adds plotly as engine to the StatsForecast's plot method.
- feat: add first version plotly plots
- feat: add engine to StatsForecast plot
- fix: legend plotly
- feat: add plotly as dependency


mergenthaler (approved):
LGTM... pun intended.

PR #315, opened 2022-11-19
[FIX] ETS for inttermitent series

FedericoGarza:
This PR fixes the ETS model for constant and inttermitent time series.

Before this change, if the series contained negative values or zeros, the multiplicative seasonal decomposition using `statsmodels` gave the error: 

`"Multiplicative seasonality is not appropriate for zero and negative values"`. 

In this PR, models with multiplicative seasonalities are omitted for these cases.

*A note on R's behavior*:

The seasonal decomposition of R actually allows doing the decomposition even though it yields `NA` values. This makes the likelihood undefined, so in the final selection these models are ignored. This PR, therefore, generates the same behavior.

- fix: make explicit machine precision np.linalg.lstsq
- fix: setting seasontype = 'N' when insufficient obs
- fix: ets numba warning of deprecation of reflection for List
- feat: add case for constant series ets
- fix: ignore multiplicative seasonality for non positive data


mergenthaler (approved):
LGTM

PR #316, opened 2022-11-19
[FIX] Theta for intermittent series

FedericoGarza:
Since the Theta methods use `seasonal_decompose`, this PR fix the multiplicative seasonality for zero and negative values.

See #315

- fix: force additive decomposition if data is non positive
- feat: protect constant time series


mergenthaler (approved):
LGTM

PR #318, opened 2022-11-21
[FEAT] Rename ETS to AutoETS

FedericoGarza:
- feat: rename ETS -> AutoETS
- fix: ets related errors
- feat: add ETS and futurewarning


mergenthaler (approved):
This change will break many examples and tutorials. I suggest you open a GitHub issue to be sure this does not affect our users.

Issue #342, opened 2022-11-29
[FEAT] make plotly visualizations scalable with plotly-resampler


jvdd:
### Description

Hi,

First of all, very cool library!! :clap: 

I think it would be interesting to make the plotly.py plots more scalable (i.e., still being responsive at 100M datapoints) by using plotly-resampler. 

The integration can be done by either;
- option A: use plotly-resampler under the hood when the `engine` in `core.GroupedArray.plot` is set to "plotly"
- option B: add "plotly-resampler" as a third option for `engine` in the `core.GroupedArray.plot` method

P.S.: I am one of the two core developers of [plotly-resampler](https://github.com/predict-idlab/plotly-resampler) and open to create a PR to add this functionality. Just want to hear (i) if you are open to this, and (ii) if so, how you would see this integration :slightly_smiling_face: 

### Use case

Having interactivity (i.e., being able to zoom, select, etc.) is crucial when analyzing large and complex data (such as time series). Plotly provides this functionality, but lacks scalability (when dealing with 10-100k+ samples) - hance plotly-resampler performs data aggregation in the back-end (your IPykernel or a seperate dash app) to update your plot dynamically.

FedericoGarza:
Hey @jvdd! Thank you! 

It sounds like a nice enhancement. 

Yes, sure! Please feel free to open a PR with your integration ❤️

PR #354, opened 2022-12-03
:sparkles: add plotly-resampler as plotting engine

jvdd:
This PR adds `"plotly-resampler"` as option for the plotting `engine` of `StatsForecast.plot` method, see #342 

Would love to hear your feedback on this!

---

Example use of plotly-resampler (in the ElectricityLoadForecasting.ipynb notetbook)

image

Note that plotly-resampler properties can be passed in the `resampler_kwargs` argument. 
:arrow_down: illustrates how for example the number of shown samples can be changed!

```py
StatsForecast.plot(df, engine="plotly-resampler", resampler_kwargs={"default_n_shown_samples": 3000, "show_dash": {"mode": "inline"}})
```

---

PS: also fixed a minor typo in the CONTRIBUTING.md file


jvdd:
I have read the CLA Document and I hereby sign the CLA

jvdd:
recheck

jvdd:
recheck

environment.yml
```diff
@@ -16,3 +16,4 @@ dependencies:
   - pip:
     - nbdev
     - tqdm
+    - plotly-resampler
```
FedericoGarza, line 19: Could you please include `plotly-resampler` as a conda dependency instead of `pip`? Thank you!

FedericoGarza (changes requested):
Thank you @jvdd! Nice integration! 

I have left a comment, once it is resolved, it will be ready to merge 🚀🤗

environment.yml
jvdd: I am fine with adding the [conda-forge feedstock](https://github.com/conda-forge/plotly-resampler-feedstock) of plotly-resampler.
However, I initially added the `pip` dependency, as 
* (@jonasvdd and me) are the maintainers of the pypi deploy - which is not the case the conda-forge deploy
* the conda-forge deploy gets updated less frequent (is still 2 versions behind the current pypi version)

So this makes me curious why this makes you prefer `conda` over `pip`? :eyes: 

P.S.: I have little to no experience with conda (for this project I created my very first conda env :upside_down_face: )

FedericoGarza: hey @jvdd! The main reason behind adding the library to conda is just as a sanity check. Since we are maintaining the [conda-forge/statsforecast](https://github.com/conda-forge/statsforecast-feedstock) repo, we must add the new dependency to the [recipe](https://github.com/conda-forge/statsforecast-feedstock/blob/main/recipe/meta.yaml) file to prevent errors or conflicts building the library. 

The alternative that comes to my mind (since you're not maintaining the conda-forge/plotly-resampler) is to try to import the library only when `engine=="plotly-resampler"` and throw an error if it is not installed.  With that change, we can include `plotly-resampler` as a core dependency in pip even if we don't include it as part of conda-forge/statsforecast. 

I'll be glad to read your thoughts on that.

jvdd: I like the alternative you suggested. I've made the necessary updates to the PR based on your suggestion.

As for the conda-forge/plotly-resampler package, I plan to become a maintainer in the future. Once I'm a maintainer, we can consider updating the dependency to conda-forge (as there will be more guarantees about the maintenance etc.).

Thanks for your feedback and suggestion!

jvdd:
I'm not sure why the [macOS - Python 3.8](https://github.com/Nixtla/statsforecast/actions/runs/3649033414/jobs/6183348089) CI is stuck at the `Run ray` step in this pull request. It looks like the same issue occurred in #358 as well.

jvdd:
Not really sure if this is related to the macOS test hanging, but GitHub is performing some maintenance on their macOS virtual environments 

https://github.com/actions/runner-images/issues/2247

FedericoGarza (approved):
Thank you @jvdd!🎉

FedericoGarza:
@all-contributors please add @jvdd for code

PR #358, opened 2022-12-06
fix: Improved notebook on prediction intervals

MMenchero:


mergenthaler:
This commit was already included in #359. Right?

Issue #287, opened 2022-10-29
Re-use parameters found by AutoARIMA


DonLeif:
I want to forward validate several auto AutoARIMA models wrapped in a StatsForecast object (multiple time series).

Is there a way to freeze once found optimal auto arima parameters to re use them (I think it's fair to assume they are relatively stable)?

I could not find such option and think it would be useful for resource-efficient repeated forecasts.

Thanks!

cwwang1128:
I just found a way to do this.

Try this:

`from statsforecast import arima
results2=arima.auto_arima_f(train.values,start_p=6,seasonal=True, start_q=5,max_p=10, max_q=10,max_order=20, period=24,allowdrift=False) #SARIMA model
print(arima.arima_string(results2,padding=False))
`
You can use this code to get the autoarima parameters, and then apply other packages, such as statsmodel, to rebuild your model. 

But it is not efficient.

Issue #94, opened 2022-04-15
Add ARIMA class


AzulGarza:
**Is your feature request related to a problem? Please describe.**
The library already has the `AutoARIMA` class but it would be helpful to have the `ARIMA` class.

**Describe the solution you'd like**
An `ARIMA` class.

PR #383, opened 2023-01-04
[FEAT] Add ARIMA model

### LINKED ISSUE #379: [Models]: provide an (non-auto)ARIMA model class as well
### Description

We'd find it useful if we could:
1. re-use the parameters found by AutoARIMA
2. sometimes, simply provide the parameters ourselves

and use your efficient implementation, instead of statsmodel's.
We think this could be achieved by providing a simple ARIMA class that we can use to manual input parameters to the ARIMA model (we do understand that there's a function we could use, but it looks like the class interface is the canonical way to access it).

Thank you!

### Use case

_No response_

FedericoGarza:
This PR fixes #379, #287, and #94.

- fix(docs): remove models numeration
- feat(models): add arima model


nelsoncardenas:
@FedericoGarza I think this PR also solves issue https://github.com/Nixtla/statsforecast/issues/406

PR #430, opened 2023-03-03
[FEAT] Add tests for ray and dask

FedericoGarza:
- feat: add ray and dask tests
- feat: recover distributed tests


FedericoGarza:
Depends on #428

Issue #436, opened 2023-03-08
Rerunning `AutoETS` on the same data gives different results


kschmaus:
### What happened + What you expected to happen

I was writing some simple tests using the `AutoETS` model, and I realized that my tests succeeded or failed based on the order they were run in. After a bit of digging, I determined that `AutoETS` learns different coefficients when trained on the same data a second time. My expectation is this would not be the case for a simple statistical model. If this lack of determinism is due to random number generation, I would expect the random seed to be exposed in the `AutoETS` model interface. 

I believe this behavior is due to the use of [mutable defaults](https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/) for `lower` and `upper` in `ets_f`. I will submit a PR to address the issue shortly. 

### Versions / Dependencies

I'm using `Python 3.10.9` and `statsforecast==1.5.0` currently. 

### Reproduction script

```python
import numpy as np
from statsforecast.models import AutoETS

np.random.seed(24601)
y = np.random.normal(size=50)

model1 = AutoETS(season_length=10, model="AAZ", damped=False)
model1.fit(y=y)
parameters1 = model1.model_["par"]

model2 = AutoETS(season_length=10, model="AAZ", damped=False)
model2.fit(y=y)
parameters2 = model2.model_["par"]

np.testing.assert_array_almost_equal(parameters1, parameters2)
# Oh no! This throws an `AssertionError`.
```

### Issue Severity

Medium: It is a significant difficulty but I can work around it.

FedericoGarza:
Solved in #437. Thank you @kschmaus! :)

PR #437, opened 2023-03-08
Stop using mutable defaults for `ets_f` `lower` and `upper` arguments

kschmaus:
This PR is meant to address Issue #436. Currently use of `ets_f` can alter the mutable defaults `lower` and `upper`, leading to non-reproducible fits of the `AutoETS` model. I can't speak to whether `lower` and `upper` should be modified over the course of a single model-fit (if they shouldn't, they could be replaced with tuples instead of arrays) - I'm just confident any modifications should be saved between model runs.


FedericoGarza:
hey @kschmaus! Thank you for solving this issue. :D

FedericoGarza:
@all-contributors please add @kschmaus for code

kschmaus:
@FedericoGarza Thank you for the great package. Happy to contribute!

PR #449, opened 2023-03-15
Added more instructions on `nbdev`

akmalsoliev:
Very little information was provided concerning the `nbdev` and its setup. Added instructions, download source and follow up commands.


FedericoGarza (approved):
Thank you @akmalsoliev! LGTM :)

PR #499, opened 2023-05-29
[FIX] Added polars to `settings.ini` 

akmalsoliev:


FedericoGarza (approved):
Thank you @akmalsoliev!

PR #511, opened 2023-06-06
[FIX] Accommodated switch to jupyter-lab

akmalsoliev:


FedericoGarza (approved):
thank you @akmalsoliev! :)

PR #557, opened 2023-06-15
[FIX] Polars documentation [skip ci]

akmalsoliev:


FedericoGarza:
Thank you @akmalsoliev!  I think that the polars tutorial you wrote should be shown in the getting started section like you suggested. Just to avoid duplicated document entries, could you remove the file `nbs/docs/how-to-guides/Getting_Started_complete_polars.ipynb`?

akmalsoliev:
> Thank you @akmalsoliev! I think that the polars tutorial you wrote should be shown in the getting started section like you suggested. Just to avoid duplicated document entries, could you remove the file `nbs/docs/how-to-guides/Getting_Started_complete_polars.ipynb` file?

sorry did not see that modifications, is there a way to update the documentation with a drop down where a user can switch between pandas and polars? 
Something like this:

kvnkho:
I was just looking around and I don't think you can do that tabs widget type view. It seems the best you can do is this: https://nbdev.fast.ai/explanations/directives.html#hide

akmalsoliev:
@kvnkho seems that that will create more confusion rather than results. Closing this PR, let's keep it as it is.

PR #558, opened 2023-06-15
[FIX] Unnecessary datetime column conversion

akmalsoliev:
Converted the column to numpy datetime even if it is already in datetime format. Unnecessary conversion.


akmalsoliev:
Env setup error, nothing major.

FedericoGarza (approved):
Thank you @akmalsoliev! LGTM :)

PR #566, opened 2023-06-22
Fix broken doc links

andrewgross:
Looks like a change in the layout of the docs broke a bunch of the links in the README, updated to point to their new locations.


kvnkho:
@all-contributors please add @andrewgross for docs

PR #571, opened 2023-06-25
[DOCS] Adding GARCH and ARCH to index

kvnkho:
Some links were still broken in Getting Started section. This is to fix that. I am also fixing the links in the index and adding the links to GARCH and ARCH models.


kvnkho:
There are two plots rendering for some reason in the notebooks. Will re-execute them.

mergenthaler (approved):
Thanks!

Issue #556, opened 2023-06-15
links to data source not working


alfonsoveneziano:
Hi
could you please update the links on the page https://nixtla.github.io/statsforecast/docs/tutorials/electricityloadforecasting.html?
For instance the link to the 'https://raw.githubusercontent.com/jnagura/Energy-consumption-prediction-analysis/master/PJM_Load_hourly.csv' is not working.

thanks a lot in advance
with regards, alfonso

kvnkho:
Thanks for reporting. Looks like they took down their repo so we'll need another source or place to host the file.

PR #572, opened 2023-06-26
Changing Load Forecasting Data Souce

kvnkho:
Fixes this: https://github.com/Nixtla/statsforecast/issues/556

Also fixes the "Getting Started" link in the nav bar which is broken.


mergenthaler (approved):
Thanks.

Issue #580, opened 2023-07-08
[Core] Exogenous Variables for Pandas DataFrame Doesn't Work


kvnkho:
### What happened + What you expected to happen

This is for the current main branch specifically, not any release.

Basically there was a snippet in `core.py`:

```python
def _parse_X_level(self, h, X, level):
    if X is not None:
        if X.index.name != "unique_id":
            X = X.set_index("unique_id")
        expected_shape_rows = h * len(self.ga)
```

and it was changed to:

```python
def _parse_X_level(self, h, X, level):
    if X is not None:
        expected_shape_rows = h * len(self.ga)
```

This now changes the comparison of `X_df` versus the expected shape, such that it thinks there is an extra column (from `unique_id` not being set as the index.

```bash
ValueError: Expected X to have shape (8, 2), but got (8, 3)
```

Script to reproduce below. I should be able to fix with just an if-else statement to handle Pandas.

### Versions / Dependencies

statsforecast main branch

### Reproduction script

```python
import pandas as pd
import numpy as np

from statsforecast import StatsForecast
from statsforecast.models import AutoARIMA

df_w_ex = pd.DataFrame(
    {
        'ds': np.hstack([np.arange(10), np.arange(10)]),
        'y': np.random.rand(20),
        'x': np.arange(20, dtype=np.float32),
    },
    index=pd.Index([0] * 10 + [1] * 10, name='unique_id'),
).reset_index(drop=False)

train_mask = df_w_ex['ds'] < 6
train_df = df_w_ex[train_mask]
print("TRAIN DF")
print(train_df.head())
test_df = df_w_ex[~train_mask]
xreg = test_df.drop(columns='y').reset_index(drop=True)
print("XREG")
print(xreg.head())
fcst_x = StatsForecast(models=[AutoARIMA()], freq='D')
res = fcst_x.forecast(df=df_w_ex[train_mask], 
                      X_df=xreg, 
                      h=4)
print(res.head())
```

### Issue Severity

Medium: It is a significant difficulty but I can work around it.

PR #581, opened 2023-07-09
Adding levels to distributed backends

kvnkho:
This PR adds levels to Spark, Dask, and Ray backends. It supports forecasting, exogenous variables, and cross validation. Tagging @goodwanghan to review also.

I tried to make the schema building more tied to the `GroupedArray` and `Statsforecast` classes. The current way in this pull request is a bit fragile since it will break if the naming convention changes. It was too invasive though to change the current set-up.

Some notes:
1. Exogenous variables in the main branch currently doesn't work for Pandas. See [this issue](https://github.com/Nixtla/statsforecast/issues/580)
2. I don't know what the cause yet and I might open another issue, but it seems the main branch `.forecast()` method returns the `unique_id` as an int even when it was a string in the original DataFrame. This probably needs some stronger guarantees and tests around.
3. Fugue needs to have ray < 2.5. This works for that, but the tests will fail if `ray>=2.5`


FedericoGarza (changes requested):
hey @kvnkho! This is really amazing!  🚀❤️

Just a minor comment: could you clean the outputs of the notebook? (`edit > clean notebooks` inside the notebook.)

Thank you!

FedericoGarza (approved):
🥳

PR #584, opened 2023-07-12
Fix incorrect parameter name in How-To-Guides

yibenhuang:
If I am correct, the parameter for defining forecast horizon in [`StatsForecast.forecast`](https://nixtla.github.io/statsforecast/src/core/core.html#statsforecast.forecast) should be `h` instead of `horizon`. 

This small fix ensures that the function call aligns with the correct usage.


kvnkho:
Thanks for the contribution @yibenhuang !

kvnkho:
@all-contributors please add @yibenhuang  for docs

PR #592, opened 2023-07-24
Conformal Prediction

kvnkho:
Some things:

1. I am least confident about the MSTL implementation if you can take a took.

2. Am I right in thinking that `prediction_intervals=ConformalIntervals(h=13, n_windows=2)` means that you need to do `forecast(h=13)`, otherwise the shapes of the intervals will not match and you can't form the output DataFrame. I feel like we should either catch this and raise an explicit error rather than a numpy shape error, or we can handle implicitly and override?

3. The type hint for level: `Optional[Tuple[int]]` causes a mypy issue because of the following behavior:
```python
test = (1, 2)
print(type(test))            # tuple
print(type(sorted(test)))    # list
```

so the line:

```
levels = sorted(levels)
```

casts tuple to list, which mypy doesn't like. So I think all the levels hints should be `Optional[List[int]]`


FedericoGarza (changes requested):
Thank you @kvnkho! This is awesome! 🥳

Could we clean the notebooks to merge the PR?🙌

kvnkho:
> Thank you @kvnkho! This is awesome! 🥳
> 
> Could we clean the notebooks to merge the PR?🙌

Finished cleaning I think.

FedericoGarza (approved):

Issue #596, opened 2023-08-02
What are the hard dependencies of statsforecast?


tim-manns:
Hello,

I'm keen to use statsforecast in AWS Lambda but the package size of 700MB is unwieldy.  Fyi - limits of 250mb for python packages are common.  I need to delete some packages and run statsforecast without many of the packages that are installed with a standard "pip install statsforecast"

I cannot find mention of the hard dependencies.   Which packages are truly needed?

Thanks 
Tim

kvnkho:
Yeah...this is related: https://github.com/Nixtla/statsforecast/issues/509

Unfortunately, I think plotly is a bit hardcoded now. It would take a bit of work to move the imports around, but definitely worth doing.

kvnkho:
I took a stab at removing plotly dependency, but the matplotlib and plotly code is not quite 1:1 yet. This will need a bit more time.

Issue #509, opened 2023-06-06
An installation method that removes visualisation dependencies such as pip install "statsforecast[core]"


DaveParr:
### Description

As discussed with @FedericoGarza a while back in https://github.com/Nixtla/statsforecast/issues/476 there could be a benefit to having some flexibility in how statsforecast is installed.

> ![](https://user-images.githubusercontent.com/8363743/236184686-dba9d26f-b5dc-4258-b37b-4ea4361e41a6.png)

> Did a mega simple analysis of the package, and by eyeball half the dependencies come from jupyter-dash, via plotly-resampler. Therefore if a version of the package can have those removed for a production api environment there should be a big simplification of environments and spin up time.

### Use case

An installation method that is more suited to server uses that contains the dependencies that are required to perform the forecasting process up to the creation of the dataframe type object, and does not include the dependencies related to plotting and visualisation.

This would reduce the burden on automated build systems and CI costs for pruduction, and the surface area of some users for due diligence and maintenance/harmonisation of dependencies.

PR #600, opened 2023-08-04
Reducing StatsForecast Size

kvnkho:
There are some issues saying that the size of StatsForecast is too inconvenient for AWS Lambda. See [this](https://github.com/Nixtla/statsforecast/issues/596) and [this](https://github.com/Nixtla/statsforecast/issues/509). The package sits at 700 MB. 

We can check the size of the dependencies below. Note that Spark, Dask, and Ray are dev dependencies. Plotly (and Dash) are core dependencies.

Screen Shot 2023-08-04 at 12 03 38 AM

This pull request removes plotly as a core dependency in favor of matplotlib being the default.

The second thing is that the build includes all source notebooks (docs and source). The docs folder is around 70-75MB. Removing the docs will also save a lot of space, especially because the docs contain plotly images.

Note:

I had to change the test to pass. It has to do with Plotly and Matplotlib not having 1:1 parity with behavior. It's somewhat ok for now because the syntax will slightly differ, but users are unlikely to use both anyway. I can open another issue for that. 

EDIT: Docs are actually not included in the build already because there is no __init__.py . Action files are but they are so small anyway it doesn't matter.


statsforecast/core.py
```diff
@@ -1686,6 +1693,33 @@ def plotly(
                 if n_cols == 1:
                     axes = np.array([axes])
 
+            def plotly(
```
jmoralez, line 1696: Why is this function being defined here?

setup.py
```diff
@@ -45,7 +46,12 @@
     packages = setuptools.find_packages(),
     include_package_data = True,
     install_requires = requirements,
-    extras_require={'dev': dev_requirements, 'dask': dask_requirements, 'ray': ray_requirements, 'spark': spark_requirements,},
+    extras_require={'dev': dev_requirements, 
+                    'dask': dask_requirements, 
+                    'ray': ray_requirements, 
+                    'spark': spark_requirements,
+                    'plotly': plotly_requirements,
+                    },
```
jmoralez, line 54:
```suggestion
    extras_require={
        'dev': dev_requirements,
        'dask': dask_requirements,
        'ray': ray_requirements,
        'spark': spark_requirements,
        'plotly': plotly_requirements,
    },
```

jmoralez (changes requested):

statsforecast/core.py
kvnkho: I really don't know. The original code that had defined there, I tried not to move it.

kvnkho: It's the equivalent here: https://github.com/Nixtla/statsforecast/blob/main/statsforecast/core.py#L1534

kvnkho: Ah, I understand! Great catch. Will update.

kvnkho: I have updated it

jmoralez (approved):
LGTM. I'm adding the breaking label because if someone is relying on statsforecast installing plotly and having it as the default backend this will require them to add it as a dependency and setting the engine argument.

kvnkho:
> LGTM. I'm adding the breaking label because if someone is relying on statsforecast installing plotly and having it as the default backend this will require them to add it as a dependency and setting the engine argument.

Yep that's right, unfortunately.

DaveParr:
Looking forward to using this, thanks for the effort. 👍🏻

nelsoncardenas:
Awesome job!

PR #601, opened 2023-08-04
Fix iteration range in non-stepwise AutoARIMA

manuel-calzolari:
Using `AutoARIMA`, I found that if you set `stepwise=False` and at least one of the `max_p`, `max_q`, `max_P` or `max_Q` parameters is set to zero, you get the following error:

```
UnboundLocalError: local variable 'best_fit' referenced before assignment
```

This is because when one of those parameters is zero, the block of code contained in these nested for loops is not executed:

```
    for i in range(max_p):
        for j in range(max_q):
            for I in range(max_P):
                for J in range(max_Q):
```

I checked the implementation of `auto.arima` in R (https://github.com/robjhyndman/forecast/blob/master/R/arima.R#L18C5-L21C0) and the corresponding code is this:

```
    for (i in 0:max.p) {
      for (j in 0:max.q) {
        for (I in 0:max.P) {
          for (J in 0:max.Q) {
```

The R loop `for (i in 0:max.p)` creates a sequence from 0 to `max.p`, inclusive. Instead, Python's `range(n)` generates a sequence from 0 to `n-1`, which is why we need to use `range(n+1)` in Python to mimic the behavior of the R loop and include `n` in the sequence. The same applies for the loops over `j`, `I`, and `J`.


kvnkho:
@MMenchero @FedericoGarza , I am positive this is right. Looking for other approvals.

kvnkho (approved):

MMenchero:
<p>I was able to replicate the issue and the error is indeed due to the range in the for loop. I have had this problem before when implementing R methods in python. I also checked the complete code with these changes and there were no other issues. </p>

kvnkho:
@all-contributors add @manuel-calzolari  for code

kvnkho:
Thanks for the contribution! Nice work! @manuel-calzolari

PR #610, opened 2023-08-15
raise informative error when series are too short for cross_validation

jmoralez:


FedericoGarza (approved):
Nice! Thank you @jmoralez. LGTM :)

Issue #602, opened 2023-08-04
Wrong shape encountered during cross validation


yuhuishi-convect:
### What happened + What you expected to happen

Encounter wrong shape when running cross validation using AutoARIMA

```
In [24]: sf.cross_validation(100, df, fitted=True)
<__array_function__ internals>:200: RuntimeWarning: invalid value encountered in cast
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-24-cd16f7527591> in <module>
----> 1 sf.cross_validation(100, df, fitted=True)

~/temp/nixtla/.venv/lib/python3.8/site-packages/statsforecast/core.py in cross_validation(self, h, df, n_windows, step_size, test_size, input_size, level, fitted, refit, sort_df)
   1515     ):
   1516         if self._is_native(df=df):
-> 1517             return super().cross_validation(
   1518                 h=h,
   1519                 df=df,

~/temp/nixtla/.venv/lib/python3.8/site-packages/statsforecast/core.py in cross_validation(self, h, df, n_windows, step_size, test_size, input_size, level, fitted, refit, sort_df)
    850         _, level = self._parse_X_level(h=h, X=None, level=level)
    851         if self.n_jobs == 1:
--> 852             res_fcsts = self.ga.cross_validation(
    853                 models=self.models,
    854                 h=h,

~/temp/nixtla/.venv/lib/python3.8/site-packages/statsforecast/core.py in cross_validation(self, models, h, test_size, fallback_model, step_size, input_size, fitted, level, refit, verbose)
    292                     else None
    293                 )
--> 294                 out[i_ts, i_window, :, 0] = y_test[:, 0] if y.ndim == 2 else y_test
    295                 if fitted:
    296                     fitted_vals[self.indptr[i_ts] : self.indptr[i_ts + 1], i_window, 0][

ValueError: could not broadcast input array from shape (41,) into shape (100,)
```

### Versions / Dependencies

```
Package                Version
---------------------- ---------
adagio                 0.2.4
ansi2html              1.8.0
antlr4-python3-runtime 4.11.1
appdirs                1.4.4
asttokens              2.2.1
backcall               0.2.0
certifi                2023.7.22
charset-normalizer     3.2.0
click                  8.1.6
contourpy              1.1.0
cycler                 0.11.0
dash                   2.11.1
dash-core-components   2.0.0
dash-html-components   2.0.0
dash-table             5.0.0
decorator              5.1.1
executing              1.2.0
[...]

PR #613, opened 2023-08-18
fix arima max order params

jmoralez:
The `max_p` and `max_q` parameters are set using the series length, which is computed using `numpy.sum` and returns a `numpy.int64`. However, there's a later check to verify that these parameters are `int`, which can fail for short series, see for example https://github.com/Nixtla/statsforecast/issues/602#issuecomment-1684335432.

This ensures that the series length is a python int, so that those parameters are python ints as well.


FedericoGarza (approved):
Thank you @jmoralez! LGTM :)

PR #614, opened 2023-08-18
add script to clean notebooks

jmoralez:
The convention for cleaning notebooks in statsforecast is:
* clear all notebooks
* for notebooks in the src directory, use the `clear_all` option (clears cell outputs as well).

This adds the `action_files/clean_nbs` script to easily replicate that and enforce it on the CI.


FedericoGarza (approved):
Really nice @jmoralez! We needed this a long time ago❤️

PR #615, opened 2023-08-21
check for level when prediction_intervals are set

### LINKED ISSUE #598: [Core] Defining Prediction Interval in StatsForecast Class Doesn't Propagate to Models

There are two places to define `ConformalIntervals`. One is that the model level, and the second one is that the `StatsForecast` level. Defining it at the `StatsForecast` level doesn't work for now. Presumably, we want to apply it to all models that have no intervals defined.

### Versions / Dependencies

Master branch (will be present in 1.6)

### Reproduction script

```python
from statsforecast.models import SimpleExponentialSmoothing, ADIDA
from statsforecast.utils import ConformalIntervals

models = [
    SimpleExponentialSmoothing(alpha=0.1),
    ADIDA()
]

res = StatsForecast(
    df=train, 
    models=models, 
    freq='H').forecast(h=24, prediction_intervals=ConformalIntervals(h=24, n_windows=2)) 

res.head().reset_index()
```

### Issue Severity

Low - there is a simply workaround

jmoralez:
Adds a check to verify that when the user provides `prediction_intervals` they also provide `level` to ensure that the method computes the intervals. For `predict` it only raises a warning because it's possible that the user only wants the mean.

Fixes #598


kvnkho (approved):

jmoralez:
Thanks @kvnkho! You should now be able to restore the example from #599 (including the level argument)

PR #617, opened 2023-08-22
take shallow copy on dataframe processing and fix get_cmap deprecation

jmoralez:
The DataFrameProcessing class always re-assigns the `ds` column, which can raise the SettingwithCopyWarning when passing subsets of pandas dataframes to it (for example in the plot function), so this takes a shallow copy before that.
Also the `matplotlib.pyplot.cm.get_cmap` function [is being deprecated](https://matplotlib.org/stable/api/prev_api_changes/api_changes_3.7.0.html#deprecation-of-top-level-cmap-registration-and-access-functions-in-mpl-cm), so this replaces it with the recommended usage [(the resampled method was added in 3.6)](https://matplotlib.org/stable/users/prev_whats_new/whats_new_3.6.0.html#colormap-method-for-creating-a-different-lookup-table-size)


FedericoGarza (approved):
Thank you @jmoralez! LGTM :)

PR #636, opened 2023-09-14
migrate plot and generate_series to utils

jmoralez:
* Changes `Statsforecast.plot` to use `plot_series` from utilsforecast.
* Changes `statsforecast.utils.generate_series` to use `utilsforecast.data.generate_series`.
* Removes the matplotlib dependency. If a user tries to use the plot method and doesn't have matplotlib installed an [useful error message is raised](https://github.com/Nixtla/utilsforecast/blob/be4382a21c4b59ec73393d1b5fba3028ac1d719d/utilsforecast/plotting.py#L15-L18)
* Deprecates the usage of 'unique_id' as the index for the `df` and `forecasts_df` argument of the plot method. Follow-up PRs will deprecate this everywhere.
* Deprecates allowing the 'ds' column to be strings, right now we convert them but in the future [an error will be raised](https://github.com/Nixtla/utilsforecast/blob/be4382a21c4b59ec73393d1b5fba3028ac1d719d/utilsforecast/validation.py#L46-L48).


FedericoGarza (changes requested):
Amazing @jmoralez! Could we update the docs that use the `plot` method to show the new plots and see if the changes integrate correctly? :)

FedericoGarza (dismissed):

Issue #629, opened 2023-08-30
Documentation on GARCH Tutorial


datatemb:
The following comment appears to be incorrect since the minimum loss value for AAPL was for the ARCH(1) model i.e. highlighted in Blue.  The comment states that GARCH(2,1) had the best result. Unless I'm just misunderstanding the tutorial, the lower the loss value the better.

See https://nixtla.github.io/statsforecast/docs/tutorials/garch_tutorial.html

![image](https://github.com/Nixtla/statsforecast/assets/58540600/1fd5ff82-8908-4bd9-9cfc-601a342a7823)

Thank you!

MMenchero:
Thanks for bringing this to our attention. It is indeed a typo and should say $ARCH(1)$. The $GARCH(2,1)$ is the best model for the next stock, which is AMZN. We'll fix it asap.

datatemb:
Thank you!


Sent from Outlook<http://aka.ms/weboutlook>
________________________________

jmoralez:
Closed by #650

PR #645, opened 2023-09-19
Fix garch model tutorial

MMenchero:
A minor issue on the tutorial for the GARCH model. This should fix issue #629. This is only a text change.


MMenchero:
Created a new PR for this issue in a clean branch.

PR #651, opened 2023-09-21
support caching all jitted functions

jmoralez:
Adds the cache and nogil arguments to all jitted functions. These can be set with:
* NIXTLA_NUMBA_CACHE: if this variable is set all functions will be cached
* NIXTLA_NUMBA_RELEASE_GIL: if this variable is set all functions release the GIL.

Also adds a small notebook (nbs/docs/how-to-guides/numba_cache.ipynb) explaining the caching.

NOTE:
The nelder_mead function couldn't be cached because it received another function as input and I couldn't figure out a way to cache it like that, so it's currently repeated for ces, ets and theta (the only difference is the objective function, which is baked into the function).


FedericoGarza (approved):
awesome @jmoralez! LGTM😍

VivCh14:
Thanks! makes sense :)

PR #656, opened 2023-10-02
migrate core processing to utilsforecast

jmoralez:
* Bumps the min python version to 3.8
* Moves polars to be an optional dependency
* Adds deprecation warnings to passing `df` to the StatsForecast constructor.
* Makes use of functions from utilsforecast to perform the processing.
* Uses the base grouped array from utilsforecast to avoid some code duplication.
* Deprecates the `sort_df` argument (in utilsforecast we sort if needed).
* Only saves the original dates and times if we'll compute the fitted values.
* Removes the warnings about ds being object and tries to parse it as it previously did.
* Removes the usage of `df` in the constructor in the fugue methods.


statsforecast/core.py
```diff
@@ -403,312 +385,24 @@ def cross_validation(
...
 
     def split_fm(self, fm, n_chunks):
         return [
-            fm[x[0] : x[-1] + 1]
-            for x in np.array_split(range(self.n_groups), n_chunks)
-            if x.size
+            fm[idxs]
+            for idxs in np.array_split(range(self.n_groups), n_chunks)
+            if idxs.size
         ]
```
AzulGarza, line 403: nice, i didn't know about this

statsforecast/core.py
```diff
@@ -838,18 +536,39 @@ def _validate_model_names(self):
...
+        self, df: Optional[DataFrame], sort_df: bool = True, save_original: bool = False
+    ) -> None:
+        if df is None:
+            _warn_df_constructor()
+            return
+        df = ensure_time_dtype(df, "ds")
+        if isinstance(df, pd.DataFrame):
+            if df.index.name == "unique_id":
+                warnings.warn(
+                    "Passing unique_id as the index is deprecated. "
```
AzulGarza, line 549: If it's deprecated, should we raise an exception instead?

statsforecast/core.py
```diff
@@ -907,49 +626,37 @@ def _make_future_df(self, h: int):
...
-            expected_shape_cols = (
-                ga_shape if not isinstance(X, pl.DataFrame) else ga_shape + 1
-            )
-            expected_shape = (expected_shape_rows, expected_shape_cols)
+        uids = np.repeat(self.uids, h)
+        if self.df_constructor is pl_DataFrame and uids.dtype.kind == "O":
+            uids = uids.astype(str)
+        df = self.df_constructor({"unique_id": uids, "ds": dates})
+        if isinstance(df, pd.DataFrame):
+            df = df.set_index("unique_id")
```
AzulGarza, line 635: for the future: should we deprecate returning an indexed data frame in the future? it would be a breaking change but more compatible with the nixtlaverse, wdyt?

statsforecast/core.py
```diff
@@ -1507,35 +1207,26 @@ def plot(
...
+            isinstance(forecasts_df, pd.DataFrame)
+            and forecasts_df.index.name == "unique_id"
+        ):
             warnings.warn(
-                "Passing an object column as 'ds' is deprecated and will raise an error in a future version. "
-                "Please convert it to datetime or integer.",
+                "Passing unique_id as the index is deprecated. "
+                "Please provide it as a column instead.",
                 category=DeprecationWarning,
             )
```
AzulGarza, line 1228: Same as before.

AzulGarza (changes requested):
This is great, @jmoralez! Abstracting `GroupedArray` in `utilsforecast` is a really cool effort. The code now is more navigable. 🎉

I just had a question regarding using `unique_id` as an index.

AzulGarza (approved):
We agreed on maintaining the `warnings` for more time to allow users to change their code. :raised_hands:

Issue #653, opened 2023-09-28
[docs] isolate import cells


jmoralez:
### Description

nbdev runs cells that contain import statements when building the docs, so when we have more code in there we incur in an unnecessary cost and it takes more time to build them, which can slow down development when previewing locally and also takes longer to deploy when pushing new commits.

We should isolate all cells that contain import statements to have only that, as is suggested [here](https://nbdev.fast.ai/explanations/docs.html#automatic-cell-execution).
> Warning
> Do not mix import statements (or calls to [show_doc](https://nbdev.fast.ai/api/showdoc.html#show_doc)) with other code in a single cell. If you do, all the code in that cell will be executed every time you build your docs, which might lead to errors (since not all previous cells will have been executed.
> Instead, put your imports in separate cells, and calls to [show_doc](https://nbdev.fast.ai/api/showdoc.html#show_doc) should contain only that one line of code – the [show_doc](https://nbdev.fast.ai/api/showdoc.html#show_doc) call.

jmoralez:
At the time of opening this issue these are the notebooks in main:

* [ ] nbs/index.ipynb
* [ ] nbs/src/adapters.prophet.ipynb
* [ ] nbs/src/mstl.ipynb
* [ ] nbs/src/distributed.utils.ipynb
* [ ] nbs/src/theta.ipynb
* [ ] nbs/src/utils.ipynb
* [ ] nbs/src/garch.ipynb
* [ ] nbs/src/ets.ipynb
* [ ] nbs/src/core/models.ipynb
* [ ] nbs/src/core/distributed.fugue.ipynb
* [ ] nbs/src/core/core.ipynb
* [ ] nbs/src/distributed.core.ipynb
* [ ] nbs/src/ces.ipynb
* [ ] nbs/src/arima.ipynb
* [ ] nbs/src/distributed.multiprocess.ipynb
* [ ] nbs/docs/how-to-guides/numba_cache.ipynb
* [ ] nbs/docs/how-to-guides/ray.ipynb
* [ ] nbs/docs/how-to-guides/AutoArima_vs_Prophet.ipynb
* [ ] nbs/docs/how-to-guides/Prophet_spark_m5.ipynb
* [ ] nbs/docs/how-to-guides/AmazonStatsForecast.ipynb
* [ ] nbs/docs/how-to-guides/ETS_ray_m5.ipynb
* [ ] nbs/docs/how-to-guides/spark.ipynb
[...]

PR #658, opened 2023-10-09
isolate import cells in distributed notebooks

jmoralez:
Contributes to #653.

Isolates the import cells in the distributed notebooks so that the docs can be built even when there are problems with the spark/java installation.


FedericoGarza (approved):
Thank you @jmoralez! LGTM :)

PR #666, opened 2023-10-13
[CHORE] Updating `requirements.txt`

akmalsoliev:
Matching `environment.yml`


dev/requirements.txt
```diff
@@ -10,8 +10,10 @@ scipy>=1.7.3
 statsmodels>=0.13.2
 tabulate
 plotly
+ray<2.5.0
```
jmoralez, line 13: I think we don't have this restriction anymore

jmoralez (changes requested):

dev/requirements.txt
akmalsoliev: @jmoralez fixed

jmoralez (approved):

PR #667, opened 2023-10-13
[FEAT] Ability to save and load StatsForecast

akmalsoliev:
methods will save and load StatsForecast class with all the models. 

Under the condition that model(s) are too big the user confirmation prompt will be asked, unless the variable `prompt` will be set to "Y"


statsforecast/core.py
```diff
@@ -1546,6 +1551,113 @@ def plot(
             palette="tab20b",
         )
 
+    def save(
+        self,
+        path: Union[os.PathLike, str],
```
jmoralez, line 1556: This doesn't seem to support pathlike objects, we can either just accept strings here or use [pathlib](https://docs.python.org/3/library/pathlib.html) inside the method

statsforecast/core.py
```diff
@@ -8,9 +8,14 @@
...
 import reprlib
 import warnings
+import os
 from os import cpu_count
 from typing import Any, List, Optional, Union, Dict
 import pkg_resources
+import pickle
+import datetime as dt
+import re
+import sys
```
jmoralez, line 18: We try to keep these organized by:
* standard library
* third party libraries
* this package (statsforecast)
so `os` should be above `warnings`, `datetime` should be the at the top, etc.

statsforecast/core.py
```diff
@@ -1546,6 +1551,113 @@ def plot(
             palette="tab20b",
         )
 
+    def save(
+        self,
+        path: Union[os.PathLike, str],
+        file_name: Union[os.PathLike, str, None] = None,
+        max_size: str = "50MB",
+        trim: bool = True,
+        trim_attr: list = ["fcst_fitted_values_", "cv_fitted_values_"],
```
jmoralez, line 1560: This is something that we should decide ourselves, it requires knowledge of the internal structures, so just having trim would be enough.
```suggestion
```

statsforecast/core.py
```diff
@@ -1546,6 +1551,113 @@ def plot(
...
+        ----------
+        path: Union[os.PathLike, str]
+            Path of the folder where the pickle should be saved to
+        file_name: Union[os.PathLike, str, None] = None
+            Name of the file, if None then datetime name will be given
+        max_size: str = "50MB"
+            StatsForecast class should not exceed this size, available byte naming:
+            ['B', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB']
+        trim: bool = True
+            Trim any boiler attributes
```
jmoralez, line 1575:
```suggestion
            Delete any attributes not needed for inference.
```

statsforecast/core.py
```diff
@@ -1546,6 +1551,113 @@ def plot(
...
+        match = re.match(r"(\d+\.\d+|\d+)(\w+)", max_size)
+        if not match or match[2] not in bytes_hmap.keys():
+            warnings.warn("Couldn't parse your max_size, default 50MB will be set")
+            cap_size = 50.0 * bytes_hmap["MB"]
+        else:
+            m_size = float(match[1])
+            key_ = match[2]
+            cap_size = m_size * bytes_hmap[key_]
+
+        sf_size = sys.getsizeof(self)
```
jmoralez, line 1610: This isn't a reliable way to get the size of the full object (the only way would be using `pickle.dumps` and getting it's size), since it doesn't look into nested attributes. Maybe we can just remove the size restriction and let the user decide if they want to keep it or not after it's been saved

statsforecast/core.py
```diff
@@ -1546,6 +1551,113 @@ def plot(
...
+            path_file = os.path.join(path, f"FittedModels_{datetime_record}.pickle")
+        else:
+            path_file = os.path.join(path, file_name)
+
+        with open(path_file, "wb") as m_file:
+            pickle.dump(self, m_file)
+        print("Model(s) saved")
+
+    @staticmethod
+    def load(path: Union[os.PathLike, str]):
```
jmoralez, line 1642: Same comment as for the save method about the pathlike annotation

jmoralez (changes requested):

statsforecast/core.py
akmalsoliev: Will add as variable inside the function.

statsforecast/core.py
akmalsoliev: fixed with `len(pickle.dumps(sf))`

statsforecast/core.py
```diff
@@ -1546,6 +1551,113 @@ def plot(
...
+        if converted_size is None or sf_byte is None:
+            error_msg = "Internal Error, this shouldn't happen, please open an issue"
+            raise NameError(error_msg)
+
+        print(f"Model(s) size: {converted_size:.2f}{sf_byte}")
+
+        print("Saving model(s)")
+
+        if not file_name:
+            datetime_record = dt.datetime.now().strftime("%m_%d_%Y_%H_%M_%S")
```
jmoralez, line 1632: We should use a more standard format here. Also using UTC time removes ambiguities about the timezone of the machine running this code.
```suggestion
            datetime_record = dt.datetime.utcnow().strftime("%Y-%m-%d_%H:%M:%S")
```

statsforecast/core.py
```diff
@@ -1546,6 +1551,113 @@ def plot(
...
+            path_file = os.path.join(path, f"FittedModels_{datetime_record}.pickle")
```
jmoralez, line 1633: I think using `pkl` is more common. Also the name should be something related to statsforecast IMO.
```suggestion
            path_file = os.path.join(path, f"StatsForecast_{datetime_record}.pkl")
```

statsforecast/core.py
```diff
@@ -1546,6 +1551,113 @@ def plot(
             palette="tab20b",
         )
 
+    def save(
+        self,
+        path: Union[Path, str],
+        file_name: Union[Path, str, None] = None,
+        max_size: str = "50MB",
```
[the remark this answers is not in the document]
akmalsoliev: how about a middle ground of if the `max_size` is set then the check will occur.

statsforecast/core.py
akmalsoliev: this format isn't allowed: https://www.mtu.edu/umc/services/websites/writing/characters-avoid/#:~:text=Illegal%20Filename%20Characters&text=Don%27t%20start%20or%20end,underscores%3B%20use%20a%20hyphen%20instead
we can do:
`"%Y_%m_%d_%H_%M_%S"` or `"%Y-%m-%d-%H-%M-%S"` or even `"%Y-%m-%d_%H-%M-%S"`

statsforecast/core.py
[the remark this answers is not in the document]
jmoralez: Sure, making it optional sounds good. But then please remove all the extra units from the dict, having up to GB seems enough.

statsforecast/core.py
jmoralez: Let's go with "%Y-%m-%d_%H-%M-%S"

akmalsoliev:
> Please also add a test in the notebook verifying that the model can be saved/loaded correctly and produces the same predictions.

Constructed a hidden cell that does that, all the modification requests are resolved in the latest push.

statsforecast/core.py
```diff
@@ -1546,6 +1551,109 @@ def plot(
...
+    ):
+        """Function that will save StatsForecast class with certain settings to make it
+        reproducible.
+
+        Parameters
+        ----------
+        path : str or pathlib.Path, optional (default=None)
+            Path of the file to be saved. If `None` will create one in the current
+            directory using the current UTC timestamp.
+        max_size: str, (default = None)
```
jmoralez, line 1568:
```suggestion
        max_size : str, optional (default = None)
```

statsforecast/core.py
```diff
@@ -1546,6 +1551,109 @@ def plot(
...
+            StatsForecast class should not exceed this size, available byte naming:
+            ['B', 'KB', 'MB', 'GB']
```
jmoralez, line 1570:
```suggestion
            StatsForecast object should not exceed this size.
            Available byte naming: ['B', 'KB', 'MB', 'GB']
```

statsforecast/core.py
```diff
@@ -1546,6 +1551,109 @@ def plot(
...
+            If max_size is set, but not parsable then default 50MB will be set.
+        trim: bool, (default = False)
```
jmoralez, line 1572:
```suggestion
        trim : bool (default = False)
```

statsforecast/core.py
```diff
@@ -1546,6 +1551,109 @@ def plot(
...
+        # @jmoralez decide future implementation
+        trim_attr: list = ["fcst_fitted_values_", "cv_fitted_values_"]
+        if trim:
+            for attr in trim_attr:
+                # remove unnecessary attributes here
+                self.__dict__.pop(attr, None)
+
+        sf_size = len(pickle.dumps(self))
+
+        if max_size:
```
jmoralez, line 1594:
```suggestion
        if max_size is not None:
```

statsforecast/core.py
```diff
@@ -1546,6 +1551,109 @@ def plot(
...
+                sf_byte = key
+                break
+
+        if converted_size is None or sf_byte is None:
+            err_messg = "Internal Error, this shouldn't happen, please open an issue"
+            raise NameError(err_messg)
+
+        print(f"Model(s) size: {converted_size:.2f}{sf_byte}")
+
+        print("Saving model(s)")
```
jmoralez, line 1614:
```suggestion
        print(f"Saving StatsForecast object of size {converted_size:.2f}{sf_byte}.")
```

statsforecast/core.py
```diff
@@ -1546,6 +1551,109 @@ def plot(
...
+
+        if not path:
```
jmoralez, line 1616:
```suggestion
        if path is None:
```

statsforecast/core.py
```diff
@@ -1546,6 +1551,109 @@ def plot(
...
+            datetime_record = dt.datetime.utcnow().strftime("%Y-%m-%d_%H-%M-%S")
+            path = f"StatsForecast_{datetime_record}.pkl"
+
+        with open(path, "wb") as m_file:
+            pickle.dump(self, m_file)
+        print("Model(s) saved")
```
jmoralez, line 1622:
```suggestion
        print("StatsForecast object saved")
```

statsforecast/core.py
```diff
@@ -8,9 +8,15 @@
 import logging
 import reprlib
 import warnings
+import errno
+import os
```
jmoralez, line 12: If we're going to use pathlib we may just use it for everything.
```suggestion
```

statsforecast/core.py
```diff
@@ -1546,6 +1551,109 @@ def plot(
...
+            Path to saved StatsForecast file.
+
+        Returns
+        -------
+        sf: StatsForecast
+            Previously saved StatsForecast
+        """
+        if not os.path.exists(path):
+            raise ValueError("Specified path does not exist, check again and retry.")
+        with open(path, "rb") as f:
```
jmoralez, line 1654: This avoids the os import and if the path is already a pathlib path then the constructor returns the same object.
```suggestion
        path = Path(path)
        if not path.exists():
            raise ValueError("Specified path does not exist, check again and retry.")
        with path.open("rb") as f:
```

jmoralez (changes requested):

statsforecast/core.py
```diff
@@ -1546,6 +1551,109 @@ def plot(
...
+
+        Parameters
+        ----------
+        path : str or pathlib.Path, optional (default=None)
+            Path of the file to be saved. If `None` will create one in the current
+            directory using the current UTC timestamp.
+        max_size: str, (default = None)
+            StatsForecast class should not exceed this size, available byte naming:
+            ['B', 'KB', 'MB', 'GB']
+            If max_size is set, but not parsable then default 50MB will be set.
```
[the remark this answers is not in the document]
akmalsoliev: `.upper()` should fix this issue

jmoralez (approved):

PR #668, opened 2023-10-13
[Feat] Robustified crossvalidation

kdgutier:
When in the presence of problematic series (eg. zero inflated) some methods like ARIMA can fail to fit.
To robustify the execution of StatsForecast.cross_validation:
- Added try/except protection to first models' fit in the cross_validation to use fall_back fitted model.
- Added forward method to Constant, Zero, and Naive models.


statsforecast/core.py
```diff
@@ -61,14 +61,20 @@ def __eq__(self, other):
...
             X = grp[:, 1:] if (grp.ndim == 2 and grp.shape[1] > 1) else None
             for i_model, model in enumerate(models):
-                new_model = model.new()
-                fm[i, i_model] = new_model.fit(y=y, X=X)
+                try:
+                    new_model = model.new()
+                    fm[i, i_model] = new_model.fit(y=y, X=X)
+                except Exception as error:
+                    if fallback_model is not None:
+                        fm[i, i_model] = fallback_model.fit(y=y, X=X)
```
jmoralez, line 75: Dont' we need to use `fallback_model.new()` here as well?

kdgutier: Thanks for the comment. I addressed it.

jmoralez (approved):

PR #694, opened 2023-11-12
Fix Minimal example and ARIMA family heading

yibenhuang:
Changes made:

- The `df` in the minimal example is not defined. I assume it is the Air Passengers data which can be accessed from the `statsforecast.utils` module.

- Update the heading level of the ARIMA family from h2 to h3, as it is supposed to be a sub heading under the "Models" section.

These two issues are both presented in the README.md and the index notebook, and they are fixed in this PR.


jmoralez (approved):
Thanks!

Issue #698, opened 2023-11-14
[Utils] `_calculate_sigma` has warning if series has 1 data point


cparmet:
Hi! Issue Severity = None because now that I understand the warning, it's no problem! Please let me know any questions. And thanks all!

#### What happened
When I fit `Naive()` models to a dataframe consisting of series with 1 point each, I get intermittent warnings:
`RuntimeWarning: invalid value encountered in double_scalars.  sigma = sigma / n`

The warning occurs once. If I fit a fresh model to the same data 50 times, the warning only occurs once. Execute the loop again, zero warnings. But restart kernel, the loop throws one warning. 

#### What I expected to happen
Forecasts looked as expected given the algorithm, and a Naive forecast on a series with one data point seems acceptable. But the warning made me look for a bug in my code or invalid data. The warning occurs when [Naive().fit()](https://github.com/Nixtla/statsforecast/blob/4fd5e0e75dbeb21eefdaf9351720cd1972789b87/statsforecast/models.py#L2856) calls [_calculate_sigma()](https://github.com/Nixtla/statsforecast/blob/4fd5e0e75dbeb21eefdaf9351720cd1972789b87/statsforecast/utils.py#L303). I could see how we might get a divide by 0 or NaN issue in a standard deviation, but the intermittent nature of the warning was confusing, since this is a deterministic model fit to the same dataset.

From [discussion with José on Nixtla Slack](https://nixtlacommunity.slack.com/archives/C05CAFFR22H/p1699980344809819):
* The issue is indeed from fitting `Naive()` to a series with only 1 data point. After the chat I understood how the `n` passed to `_calculate_sigma()` is the series length - 1, so with my dataset `n=0`, and `sigma / n` divides by zero, throwing the warning.
* The intermittency appears to be related to the `warnings` module, but the divide by zero occurs every time.

José's suggestion:

> We should probably change that so that if n is zero sigma is 0 or similar. Can you open an issue for that?

### Versions / Dependencies

statsforecast 1.5.0 and 1.6.0
A Python 3.8 image on SageMaker

Also reproduced in Colab with Python 3.10.12, statsforecast 1.6.0

### Reproduction script

```python
# !pip install pandas statsforecast==1.6.0 # if running in notebook
import pandas as pd
from statsforecast import StatsForecast
from statsforecast.models import Naive

X = pd.DataFrame({'unique_id':[0,1,2,3,4], "ds":[1,1,1,1,1], "y":[10,20,30,40,50]})

[...]

PR #699, opened 2023-11-14
feat: _calculate_sigma() explicitly handles series with 1 data point (n=0)

cparmet:
Addresses #698

An internal warning was raised if a `Naive()` model is fit to a series with only 1 data point. That's because calculating `sigma` involves a divide by zero. This PR handles that scenario with the method suggested by @jmoralez.

Note: The [CONTRIBUTING.md](https://github.com/cparmet/statsforecast/blob/main/CONTRIBUTING.md) guide refers to `./action-files/lint`. I don't see this file in `main`. Did I miss it or is there another way to perform the desired linting used by `statsforecast`?


jmoralez (approved):
Thanks!

PR #701, opened 2023-11-15
add mstl_decomposition function

jmoralez:
Adds the feature_engineering module which will hold utilities to help generate features to be used in downstream models. Right now we're adding the `mstl_decomposition` function, which adds the trend and season component of each serie as well as a guide on how to use it.


AzulGarza:
I'm loving this already😍

AzulGarza (approved):
nice!

Issue #690, opened 2023-11-09
[ARIMA] Incorrect computation of n 


jmoralez:
### What happened + What you expected to happen

`n` here won't have the number of non NA values, but the number minus 1, because the end of the slice isn't included. This was originally reported in #654 since if there's a single sample then `n=0` and then we try to take its log.
https://github.com/Nixtla/statsforecast/blob/be9db750ab1fc858b6810f1395f36895a9e2f447/statsforecast/arima.py#L1212-L1214

### Versions / Dependencies

1.6.0

### Reproduction script

```python
import numpy as np
from statsforecast.models import AutoARIMA

AutoARIMA().fit(np.array([1.0]))
```

### Issue Severity

None

jmoralez:
Hey @StatKumar. Sure, go ahead! We have a contributing guide [here](https://github.com/Nixtla/statsforecast/blob/main/CONTRIBUTING.md) but feel free to ask any questions.

We also have the bug here: https://github.com/Nixtla/statsforecast/blob/be9db750ab1fc858b6810f1395f36895a9e2f447/statsforecast/arima.py#L1483-L1485
and here: https://github.com/Nixtla/statsforecast/blob/be9db750ab1fc858b6810f1395f36895a9e2f447/statsforecast/arima.py#L1815-L1817
The fix would be adding a +1 to the `lastnonmiss`, i.e. `lastnonmiss = missing_idxs.max() + 1` so that the last index is considered in the slice below.

StatKumar:
Thank you @jmoralez

PR #708, opened 2023-11-24
issue:fix for issue #690 - incorrect computation of n

StatKumar:
This is the fix for the issue # 690. 

Math error was occurring when the len(n) = 1. While slicing the array, the last value of the array is not considered, and due to slicing, a variable's value was set to 0, and later in the code, log of that variable was taken, which led to math error. 

Fix:
n = np.sum(~missing[firstnonmiss:lastnonmiss]) 
changed to
n = np.sum(~missing[firstnonmiss:])


jmoralez:
Hey @StatKumar. Can you please:
* Read the [contributing guidelines](https://github.com/Nixtla/statsforecast/blob/main/CONTRIBUTING.md) and make the change in the notebook.
* Make the change I suggested [here](https://github.com/Nixtla/statsforecast/issues/690#issuecomment-1824642237) (`lastnonmiss = missing_idxs.max() + 1`). The proposed change here would unnecessarily break the code for people who are directly using the AutoARIMA model and passing arrays like [nan, 1, 2, 3, nan].
* Also change the other places where this is happening, which I referenced [here](https://github.com/Nixtla/statsforecast/issues/690#issuecomment-1824642237) as well.

StatKumar:
@jmoralez Thank you for the guidelines. Will update soon.

jmoralez:
Can you update this PR instead of opening a new one?

StatKumar:
Sure @jmoralez

StatKumar:
@jmoralez Updated the pull request with a new commit. Please let me know if anything else needs to be done. Thank you for taking time in guiding me through my first issue in statsforecast.

jmoralez:
Thanks @StatKumar! This also happens in the `myarima` function. Can you add the fix there as well? Also please add a cell at the bottom with the original example from #690 (`AutoARIMA().fit(np.array([1.0]))`) to make sure that the fix works.

jmoralez (approved):

PR #715, opened 2023-11-29
improve deprecation error messages

jmoralez:
The deprecation message for not providing a df to one of the methods that needs it always referenced the constructor, but they'll also be printed if someone reuses a dataset built with another method, e.g. the following code
```python
sf = StatsForecast(models, freq)
sf.fit(df=df)
sf.forecast(h=2)
```
would show the deprecation warning about providing the df to the constructor, even though it wasn't provided to it. This changes the message slightly to suggest that the df may come from either the constructor or a previous method.

Also now raises an informative error if the df wasn't provided at all, e.g. the following code
```python
sf = StatsForecast(models, freq)
sf.fit()
```
would raise a confusing error about the StatsForecast object not having a `ga` attribute. This now asks for the `df` argument to be provided.

### DIFF
--- a/nbs/src/core/core.ipynb
+++ b/nbs/src/core/core.ipynb
@@ -838,9 +838,9 @@
     "#| exporti\n",
     "def _warn_df_constructor():\n",
     "    warnings.warn(\n",
-    "        \"The `df` argument of the StatsForecast constructor is deprecated \"\n",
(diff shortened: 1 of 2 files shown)


AzulGarza (approved):
LGTM :)

PR #718, opened 2023-12-05
support different column names for ids, times and targets

jmoralez:
* Adds support for providing `id_col`, `time_col` and `target_col` to the methods that take a dataframe so that different column names can be used, i.e. removes the restriction on unique_id, ds, y.
* Makes the forwarding of the arguments to the distributed backends explicit.
* Removes the distributed.utils module, which doesn't seem to be used anymore along with the _nbdev.py file (from nbdev1).
* ~~Fixes some docstrings but I'll make another PR to define them only once and set them in all of the methods to make it easier and ensure they're consistent.~~ Templates the common arguments in the docstrings to ensure consistency and to make it easier to update them.


statsforecast/_nbdev.py
AzulGarza, line 1: Thanks for removing this file.

statsforecast/core.py
```diff
@@ -678,10 +703,9 @@ def fit_predict(
...
             Forecast horizon.
-        df : pandas.DataFrame | polars.DataFrame, optional (default=None)
-            DataFrame with columns [`unique_id`, `ds`, `y`] and exogenous variables.
-            If None, the `StatsForecast` class should have been instantiated
-            using `df`.
+        df : pandas or polars DataFrame, optional (default=None)
+            DataFrame with columns [`id_col`, `time_col`, `target_col`] and exogenous.
+            If None, the `StatsForecast` class should have been instantiated using `df`.
         X_df : pandas.DataFrame | polars.DataFrame, optional (default=None)
             DataFrame with [`unique_id`, `ds`] columns and `df`'s future exogenous.
```
AzulGarza, line 710: Should we also update the `X_df` documentation?

statsforecast/core.py
```diff
@@ -1369,8 +1510,8 @@ class StatsForecast(_StatsForecast):
         See [panda's available frequencies](https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases).
     n_jobs : int (default=1)
         Number of jobs used in the parallel processing, use -1 for all cores.
-    df : pandas.DataFrame | pl.DataFrame, optional (default=None)
-        DataFrame with columns [`unique_id`, `ds`, `y`] and exogenous.
+    df : pandas or polars DataFrame, optional (default=None)
+        DataFrame with columns [`id_col`, `time_col`, `target_col`] and exogenous.
     sort_df : bool (default=True)
         If True, sort `df` by [`unique_id`,`ds`].
```
AzulGarza, line 1516: "If True, sort `df` by id and time."

AzulGarza (changes requested):
Thank you, @jmoralez! This is a great improvement. I just left a comment regarding the documentation of `X_df`.

statsforecast/core.py
jmoralez: Added the templating in 678f3c1

AzulGarza (approved):
🎉

PR #720, opened 2023-12-06
remove spaces in single-line equations and place tags inside equation context

jmoralez:
* Removes spaces after opening and before closing $, e.g. `$ \alpha$` becomes `$\alpha$`
* Places equations with the `\tag` command inside an equation context.
* Adds the script that was used to perform these replacements.


mergenthaler (dismissed):
Thanks for the work.

PR #747, opened 2024-01-04
fix arima init params

jmoralez:


statsforecast/arima.py
```diff
@@ -1232,11 +1229,9 @@ def myarima(
             fit["coef"] = change_drift_name(fit["coef"])
         else:
             if use_season:
-                fit = arima(
-                    x, order, seasonal, include_mean=constant, method=method, xreg=xreg
-                )
+                fit = arima(x, order, seasonal, method=method, xreg=xreg)
             else:
-                fit = arima(x, order, include_mean=constant, method=method, xreg=xreg)
+                fit = arima(x, order, method=method, xreg=xreg)
```
AzulGarza, line 1234: I think this part refers to [here](https://github.com/robjhyndman/forecast/blob/681359b1afb8a117944ca165c98dcfd594861e59/R/newarima2.R#L711-L717) (which includes `include_mean=constant`)  instead of [here](
https://github.com/robjhyndman/forecast/blob/681359b1afb8a117944ca165c98dcfd594861e59/R/newarima2.R#L705-L709).

AzulGarza (changes requested):
thank you @jmoralez! 🥳

i only have one comment regarding removing `include_mean`.

statsforecast/arima.py
jmoralez: Restored in ee349da

AzulGarza (approved):
🥳

Issue #751, opened 2024-01-08
Issue calculating conformity scores


jcatankard:
Hi - I am not sure if this is an error or I am misunderstanding the implementation of conformity scores. Hopefully you can shine some light for me.

X (exogenous regressors) is not utilized in method at all, instead X_train and X_future are derived from y. Is this correct?

```
X_train = y[:, 1:] if (y.ndim == 2 and y.shape[1] > 1) else None
y_test = y[cutoff:] if end_cutoff == 0 else y[cutoff:end_cutoff]
X_future = (
    y_test[:, 1:] if (y_test.ndim == 2 and y_test.shape[1] > 1) else None
)
```
https://github.com/Nixtla/statsforecast/blob/3b642177351ab802621d662daaca2bbdacba680e/statsforecast/models.py#L129

Also, is it not always expected that `y.ndim == 1` if y is a one dimensional timeseries? As such, X_train and X_future always resolve to None and then no regressors are used for forecasting.

jcatankard:
FYI, I resolved my own code by replacing the the for loop with:
```
for i_window, cutoff in enumerate(steps, start=0):
    end_cutoff = cutoff + h

    if X is None:
        X_train = None
        X_future = None
    else:
        X_train = X[:cutoff, :]
        X_future = X[cutoff:, :] if end_cutoff == 0 else X[cutoff: end_cutoff, :]

    y_train = y[:cutoff]
    y_test = y[cutoff:] if end_cutoff == 0 else y[cutoff: end_cutoff]

    fcst_window = self.forecast(h=h, y=y_train, X=X_train, X_future=X_future)  # type: ignore[attr-defined]
    cs[i_window] = np.abs(fcst_window["mean"] - y_test)
```

jmoralez:
Hey. I believe y contains X_train and we should the X argument as the future. Are you interested in submitting a PR?

jcatankard:
Hi again. I made a PR here: https://github.com/Nixtla/statsforecast/pull/755
Thanks!

PR #755, opened 2024-01-09
Issue/751: using X for calculating prediction intervals

jcatankard:
Resolving https://github.com/Nixtla/statsforecast/issues/751

X (exogenous regressors) can now be used to create prediction intervals if they are passed to the function separately from y. Now, if y is one dimensional, it will use X instead to create X_train and X_future.

Also, evaluate y.ndim outside of the for loop so it only has to be checked once.

### DIFF
--- a/statsforecast/models.py
+++ b/statsforecast/models.py
@@ -117,6 +117,9 @@ class _TS:
         y: np.ndarray,
         X: Optional[np.ndarray] = None,
     ) -> np.ndarray:
+        if y.ndim == 2:
+            X = y[:, 1:] if y.shape[1] > 1 else X
+            y = y[:, 0]
         n_windows = self.prediction_intervals.n_windows  # type: ignore[attr-defined]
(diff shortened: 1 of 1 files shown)


jmoralez:
Hey @jcatankard, thanks for the fix. I was wrong in https://github.com/Nixtla/statsforecast/issues/751#issuecomment-1881866850, we split `X` and `y` here: https://github.com/Nixtla/statsforecast/blob/713403f5734d14bc882c579b8c3e0a74469ab64b/statsforecast/core.py#L54-L55
so in the models `y` is a 1d array and `X` is either a 2d array or `None`. Can you please update this PR? It should always slice `y` and only slice `X` when it is not `None`.

jcatankard:
Thanks @jmoralez for coming back around to this. I see you made some commits, do you need anything else from me?

AzulGarza (approved):
thank you @jcatankard for the initial effort and @jmoralez for the additional changes. ❤️ 
LGTM

Issue #753, opened 2024-01-09
remove numba dependency 


jmoralez:
### Description

We've heavily relied on [numba](https://numba.pydata.org/) to speed up our models, however we don't have the need for its JIT compilation, since the code that uses it is defined inside the library.

Replacing numba jitted code with a compiled alternative (C++ or Rust for example) would provide the following benefits:
* Less dependencies, so less friction and also smaller environments (LLVM is a bit heavy).
* Reducing cold starts, e.g. #162, #177.
* Not having to wait for support of new features, like https://github.com/Nixtla/statsforecast/issues/424#issuecomment-1498425990.

And the following drawbacks:
* More time to develop new models.
* It would make it more difficult for users to install from github since they'd need to have a compiler installed. Although we could publish nightly wheels for this.

### Use case

This will benefit the development process since even when using the cache it can take a couple of seconds to run jitted functions for the first time.
Also deployments would be smoother because either:
1. People that were copying over the cache won't have to do it anymore.
2. People that weren't copying the cache won't experience cold starts anymore.

AzulGarza:
i love this. i think we should move forward with the enhancement. 

regarding the second drawback, i agree with the nightly wheels option. also adopting a compiled alternative will force us to make releases more often. 🙌

regarding the first drawback, we could still have some models relying on numba as an optional dependency. for example, if we release a new model written in numba, it will be only available installing `statsforecast[numba]`. this approach could help us iterate new models faster and once they are stable, we could migrate them to their compiled version. :)

jmoralez:
I used the following to profile the current compilation times:

<details><summary>Click to expand</summary>

```python
import datetime
import operator
import os
os.environ.pop('NIXTLA_NUMBA_CACHE', None)
from collections import defaultdict

import numpy as np
import pandas as pd
from numba.core.event import install_recorder

from statsforecast.core import StatsForecast
from statsforecast.models import *
from statsforecast.utils import generate_series

data = generate_series(2)
models = [
    AutoARIMA(season_length=7),
    AutoCES(season_length=7),
    AutoETS(season_length=7),
    AutoTheta(season_length=7),    
    SimpleExponentialSmoothing(alpha=0.1),
    GARCH(),
    TBATS(seasonal_periods=7),
]
sf = StatsForecast(models=models, freq='D')
with install_recorder("numba:compile") as rec:
    forecast = sf.forecast(df=data, h=7)

events = defaultdict(dict)
for ts, event in rec.buffer:
[...]

PR #757, opened 2024-01-12
migrate ETS to C++

jmoralez:
Contributes to #753 by moving the functions that used numba in the ETS model to C++. Several changes had to be made to the library for this to work:

* Move the python package to a python folder. This is so that when running the tests from the root of the repository it uses the installed statsforecast package instead of the code from the statsforecast folder. With a regular python package these would be the same, however since we're now shipping a shared library these two are now different (the installed package has the shared library and the statsforecast folder doesn't).
* The tests now build the wheels first for the platforms that we'll support (windows amd64, macos x864_64, macos arm64, linux x86_64, linux aarch64 (arm64)) and then run the tests installing from that. These wheels will be published to s3 on pushes to main so that people can install from those without needing a compiler (the install notebook has instructions on how to do this).
* Updated the python-publish action to build the wheels for the same platforms as in the previous point.
* Added pybind11 to the dev requirements, which is used to build the shared library.
* Moved the restrict_to_bounds function and the results tuple from ETS to utils. These will be removed once we migrate all the models that use numba to C++.


jmoralez:
Pinning pandas version because of [this](https://discuss.ray.io/t/pandas-importerror-with-ray-data-dataset-show/13486).

AzulGarza (commented):
awesome @jmoralez! thank you :)

just a minor comment, could you update the `CONTRIBUTING.md` file to reflect these changes? i think in this case it should be particular to statsforecast (for example: how to add a new model written in C++).❤️

jmoralez:
@AzulGarza I think this is ready now. The changes since you last reviewed are:

* Switched to [pybind11](https://github.com/pybind/pybind11) to generate the python bindings. The main benefits are:
    * The overhead of calling a function is very low, which is great for objective functions that get called during numerical optimization.
    * The interchange between the two languages is easier, e.g. you can define that you want a double array and pybind11 will try to coerce the input to that if it isn't and raise an error otherwise, you can return tuples, etc.

    The downside is that we now have to build wheels for each python version instead of just one per os-architecture combination.

* Restored the setup.py file and extended it to build the shared library. This is so that we don't have to learn CMake, we just use that script.

* Updated the CONTRIBUTING.md file to use uv (since I deleted the conda environment files in #874 to have all dependencies in a single place) and added a section on how to rebuild the extension (I can extend on that if you want but it's just a matter of changing/creating the C++ files and rebuilding).

AzulGarza (approved):
this is amazing, @jmoralez! congrats🎉 lgtm

PR #766, opened 2024-01-22
MSTL fixes

### LINKED ISSUE #675: MSTL forecast result different compared to R's forecast::mstl()

As I understand it, statsforecast's `MSTL` aims at implementing R's `forecast::mstl` function in Python. However, when comparing forecasts on different datasets I always end up with a different result. Looking in the documentation I can't identify any parameter defaults that are different in `statsforecast.MSTL` vs `forecast::mstl`.

See reproduction scripts below.

### Versions / Dependencies

statsforecast v 1.6.0

### Reproduction script

R `forecast::mstl()` :
```
library(data.table)
library(magrittr)
hufl = fread("https://raw.githubusercontent.com/zhouhaoyi/ETDataset/main/ETT-small/ETTm1.csv") %>% 
  .[, HUFL]

hufl %>% 
  forecast::msts(seasonal.periods = c(24 * 4, 4)) %>% 
  forecast::mstl() %>% 
  predict(h = 1) %>% 
  .[["mean"]]

# [1] 12.17458

```

Python `statsforecast.MSTL()`:
```
import pandas as pd
from statsforecast.models import MSTL

mstl_model=MSTL(season_length=[24 * 4, 4])
hufl=pd.read_csv("https://raw.githubusercontent.com/zhouhaoyi/ETDataset/main/ETT-small/ETTm1.csv")["HUFL"].to_numpy()
mstl_model.forecast(hufl, 1)
# {'mean': array([9.76441481])}
```

### Issue Severity

High: It blocks me from completing my task.

jmoralez:
Does the following changes to be closer to the R version:
* Sets two iterations as the default and resets to 1 if there's only one seasonality.
* Sets STL's seasonal_deg=0 to be equal to the s.degree in the [R version](https://www.rdocumentation.org/packages/stats/versions/3.6.2/topics/stl).
* Sorts the seasonalities in ascending order.

Fixes #675


statsforecast/mstl.py
```diff
@@ -15,16 +15,20 @@ def mstl(
...
     origx = x
     n = len(x)
-    msts = [period] if isinstance(period, int) else period
-    iterate = 1
+    if isinstance(period, int):
+        msts = [period]
+    else:
+        msts = sorted(period)
+    if len(msts) == 1:
+        period = 1
```
AzulGarza, line 31: i think this part is unnecessary (we don't use `period` after defining `msts`). non-blocking comment. :)

AzulGarza (dismissed):
LGTM❤️

statsforecast/mstl.py
jmoralez: That should be `iterate`. I forgot to export but the check passed, I'll investigate that.

AzulGarza (approved):
thanks!🚀

PR #768, opened 2024-01-23
check exogenous features are present when needed

### LINKED ISSUE #618: AutoARIMA crashs if unexpected column in dataframe

If the dataframe used to train AutoARIMA contains an additional column not just the expected unique_id | ds | y it crashes during sf.predict()

I would expect it to ignore the additional col.

### Versions / Dependencies

Colab

### Reproduction script
```python
import pandas as pd
from statsforecast.utils import AirPassengersDF

df_all = AirPassengersDF
# Extract the month from the 'ds' column and add it as a new column
df_all['ds'] = pd.to_datetime(df_all['ds'])
df_all['month'] = df_all['ds'].dt.month # comment this line to make code run

sf = StatsForecast(
    models = [AutoARIMA(season_length = 12)],
    freq = 'M'
)
sf.fit(df_all)
sf.predict(h=12)
```

### Issue Severity

Low: It annoys or frustrates me.

jmoralez:
Saves the exogenous features used during training to provide an informative error if they're missing on the predict step and they're required by a model.

Fixes #618


AzulGarza (approved):
nice!

PR #788, opened 2024-02-23
tbats fixes

jmoralez:
Makes the following bug fixes:
* Applies the transformation with the optimum lambda to the seed states when using boxcox.
* Removes the Augmented Dickey-Fuller test.
* Checks if `use_trend in (True, None)` to correctly generate the combinations in the auto model.
* Renames `seasonal_periods` to `season_length` to be consistent with the other models.

Also modifies the process to find the optimal number of harmonics by:
* Using an expanding mean if there aren't enough samples in the window to remove the trend (thus doesn't produce missing values). We discussed if we should use `center=True` and found that overall the forecasting error is smaller when using `center=False`.
* Subtracting the signal found for each seasonal period before looking for the next one

Adds an experiment with the M4 hourly dataset.


nbs/src/tbats.ipynb
```diff
@@ -90,56 +90,50 @@
     "\n",
```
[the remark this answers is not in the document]
jmoralez: We discussed using `.rolling(window=window_size, min_periods=1, center=True)` here but it seems that we get a lower forecasting error by using `center=False` (the default), so we're leaving it with the default for now.

MMenchero (approved):
These are small but key fixes to the code.

AzulGarza (dismissed):

nbs/src/tbats.ipynb
```diff
@@ -894,7 +890,18 @@
     "        log_likelihood = len(y_trans)*np.log(np.nansum(errors**2))\n",
     "\n",
     "    kval = len(optim_params)+x_nought.shape[0]\n",
-    "    aic = log_likelihood+2*kval \n",
+    "    \n",
```
MMenchero, line 893: I found that some time series "explode" (for example, Y21480 from the M4 dataset). 
This change in how the AIC is computed helps fix some of these cases. 
This was proposed by [De Livera, Hyndman and Snyder](https://citeseerx.ist.psu.edu/document?repid=rep1&type=pdf&doi=f3de25596ab60ef0e886366826bf58a02b35a44f) on page 20.

nbs/src/tbats.ipynb
```diff
@@ -1005,34 +1013,40 @@
...
     "\n",
     "    # Select combinations to try out \n",
     "    if use_boxcox is None: \n",
     "        B = [True, False]\n",
     "    else: \n",
     "        B = [use_boxcox]\n",
     "\n",
-    "    if use_trend: \n",
-    "        if use_damped_trend is None: \n",
+    "    if use_trend is None:\n",
```
MMenchero, line 1034: Some models were undefined. For example, when using `use_trend=True` and `use_damped_trend=None`, one of the resulting combinations was [True, None]. As a result, not all possible models were being considered.

nbs/src/core/models.ipynb
```diff
@@ -9951,6 +9960,8 @@
     "                res = {**res, **fitted_pred_int}\n",
     "        if mod['BoxCox_lambda'] is not None:\n",
     "            res_trans = {k : inv_boxcox(v, mod['BoxCox_lambda']) for k, v in res.items()}\n",
+    "            for k, v in res_trans.items():\n",
```
MMenchero, line 9963: I found that in some cases the [inverse Box-Cox ](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.inv_boxcox.html)generates missing values. One example is series Y11663 from the M4 dataset. This is because there is no x such that $y = (x^\lambda - 1) / \lambda$ for $\lambda \neq 0$. For now, I just returned the non-transformed value.

experiments/tbats/README.md
```diff
@@ -1,70 +1,89 @@
 # TBATS: A brand-new implementation for the Python ecosystem
 
-## TL;DR 
-For next-day forecasting of electricity load, StatsForecast's new implementation of TBATS outperforms current implementations of this model in both accuracy and speed.
+## TL;DR
```
MMenchero, line 3: I think results are ok for the M3 dataset since we're close to R's. Yearly performs worse than the seasonal naive, but R is no better. The results for the M4 dataset suggests we should keep checking the code, in particular for hourly. 

I removed the original dataset (electricity) because results got worse after fixing a couple of bugs in the previous version.

MMenchero (changes requested):

MMenchero:
Important changes made by @jmoralez:
- Uses a new Box-Cox transformation that fixes the cases where the inverse transformation results in missing values. 
- Limits the number of searches when determining the number of harmonics.
- Adds a scale parameter to the objective function, akin to what R does. 
With these changes, I reran the complete experiments for the M3 and M4 datasets.

MMenchero (dismissed):

AzulGarza (commented):
very nice work  @jmoralez @MMenchero! LGTM

PR #790, opened 2024-02-24
fix: parallel custom cols

AzulGarza:
this PR fixes the following bug: when using a custom target column in parallel mode (`n_jobs` and `n_series` greater than 1), the `forecast_fitted_values()` method returns a dataframe with columns `["<target_col>", "<time_col>", "y"]` independently of the value of `target_col`.  additionally, the bug caused always printing the progress bar while using the  `forecast` and `cross_validation` methods in parallel mode.  the fix also allows enabling or disabling the progress bar in parallel mode. a test pytest was included to verify the solution works in a parallel environment.


action_files/test_fit_predict.py
```diff
@@ -4,22 +4,74 @@
...
+            "y": "target",
+        }
+        panel_df.rename(columns=renamer, inplace=True)
+    return panel_df
+
 
-@pytest.mark.parametrize('n_jobs', [-1, 1])
+@pytest.mark.parametrize("n_jobs", [-1, 1])
+def test_custom_cols(n_jobs):
+    AirPassengersPanel = get_data(rename_cols=True)
```
VascoSch92, line 46: why `PascalCase` if it is an instance of a pandas data frame? are `PascalCase` not used just for class names?

jmoralez:
Thanks! Given that the `apply_async` [supports keyword arguments](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool.apply_async) I think it might be better to explicitly list them as we do here https://github.com/Nixtla/statsforecast/blob/b61476e603e1fe7a681fee2ad3254130962ce1b7/statsforecast/distributed/fugue.py#L110-L120 to avoid breaking this again. WDYT?

jmoralez (approved):

PR #828, opened 2024-04-24
Fix allowdrift and allowmean in non-stepwise AutoARIMA

manuel-calzolari:
Non-stepwise AutoARIMA calls the `search_arima` function by passing the arguments `allowdrift=allowdrift` and `allowmean=allowmean`. However, in the `search_arima` function, the correct parameters are `allow_drift` and `allow_mean` (with underscores). As a result, for non-stepwise AutoARIMA, the default values of the `search_arima` function, which are `allow_drift=True` and `allow_mean=True`, are always used. Thus, passing, for example, `allowdrift=False` to non-stepwise AutoARIMA has no effect. This patch corrects the names of the parameters being passed.


manuel-calzolari:
A more intrusive alternative might be to rename `allow_drift` and `allow_mean` to `allowdrift` and `allowmean` in the `search_arima` function, but this might break backward compatibility if someone calls the `search_arima` function directly.

jmoralez:
Thanks @manuel-calzolari, this makes sense. Can you please change `macos-latest` here: https://github.com/Nixtla/statsforecast/blob/057cacd7b2e14572d32c3483d5ff5ef2cfd539a0/.github/workflows/ci.yaml#L74
to `macos-13`? Seems like now `latest` points to the M1 runners and prophet isn't built for arm64 in conda-forge.

jmoralez (approved):
Thanks!

PR #895, opened 2024-09-03
feat: migrate arima to c++

jmoralez:


experiments/m3/src/evaluation.py
```diff
@@ -50,11 +50,11 @@ def main(test: bool = False):
     time = evaluation.query('metric=="time"').T
     if test:
         expected_results = {
-            'AutoARIMA': 4.87, 
+            'AutoARIMA': 4.57,
```
jmoralez, line 53: there were a couple of bugs that impacted the performance, so fixing them reduced this metric but slightly increased the ensemble one (my guess is that some of the errors were actually helping the ensemble)

python/statsforecast/arima.py
```diff
@@ -950,13 +512,16 @@ def arma_css_op(p, x):
...
-            init = ARIMA_invtrans(init, arma)
-            if arma[1] > 0:
-                ind = arma[0] + np.arange(arma[1])
-                init[ind] = maInvert(init[ind])
-            if arma[3] > 0:
-                ind = np.sum(arma[:3]) + np.arange(arma[3])
-                init[ind] = maInvert(init[ind])
+            candidate = ARIMA_invtrans(init, arma)
+            phi, _ = arima_transpar(candidate, arma, transform_pars)
+            if np.logical_and(phi > -math.pi / 2, phi < math.pi / 2).all():
```
jmoralez, line 517: The invtrans is one of the functions that had a bug, so with the bug fixed this can sometimes generate invalid init states and we have to check them.

setup.py
```diff
@@ -61,6 +61,7 @@
         cxx_std=17,
     )
 ]
+ParallelCompile("CMAKE_BUILD_PARALLEL_LEVEL", needs_recompile=naive_recompile).install()
```
jmoralez, line 64: This only re-compiles *.cpp files that have changed since the last compilation.

experiments/m3/src/evaluation.py
AzulGarza: cool

python/statsforecast/arima.py
AzulGarza: nice

AzulGarza (approved):

Issue #897, opened 2024-09-05
[AutoARIMA] Major hidden bug in `auto_arima_f` 


robert-robison:
### What happened + What you expected to happen

A bug in `auto_arima_f` is causing the wrong model to be returned in many cases. What frequently is returned is the best model but with no constant terms.

The relevant section is in the `try_params` function within `auto_arima_f`: https://github.com/Nixtla/statsforecast/blob/7f60571b1242413b372028504e60b56d0d566214/statsforecast/arima.py#L2149C1-L2152C10

The issue the `constant` value is not being passed through to `p_myarima`, so even if you set `constant = False` in the `try_params` call, the arima model that fits will still include it.

This is the chain reaction that leads to the wrong model being selected:

1. Toggling the `constant` value is the last part of the stepwise while loop. If we get there, all other options have been exhausted.
2. Since the `constant` toggle is being ignored in `try_params`, it will return the same `fit["ic"]` value as the `bestfit` model. Since it's the exact same, `improved` will be `False`, and we will exit the stepwise process.
3. The code immediately after the stepwise process uses `np.argsort` to sort by lowest ic value. There will be two models with the same ic--the actual best model, and the best model with `constant = False`. In my experiments, for whatever reason it often selects the latter.
4. The result is the actual best model with the constant removed, which is often significantly worse.

The current code:
```python
    def try_params(p, d, q, P, D, Q, constant, k, bestfit):
        improved = False
        if k >= results.shape[0]:
            return k, bestfit, improved
        fit = p_myarima(
            order=(p, d, q),
            seasonal={"order": (P, D, Q), "period": m},
        )
        results[k] = (p, d, q, P, D, Q, constant, fit["ic"])
        k += 1
        if fit["ic"] < bestfit["ic"]:
            bestfit = fit
            improved = True
        return k, bestfit, improved
```

What the code should be:
```python
    def try_params(p, d, q, P, D, Q, constant, k, bestfit):
        improved = False
        if k >= results.shape[0]:
            return k, bestfit, improved
        fit = p_myarima(
            order=(p, d, q),
            seasonal={"order": (P, D, Q), "period": m},
            constant=constant,  # This is the line to be added
        )
        results[k] = (p, d, q, P, D, Q, constant, fit["ic"])
        k += 1
        if fit["ic"] < bestfit["ic"]:
            bestfit = fit
            improved = True
        return k, bestfit, improved
```

### Versions / Dependencies

<details><summary>Click to expand</summary>
[...]

PR #899, opened 2024-09-06
fix: pass constant in autoarima stepwise

robert-robison:
Fix a bug in `auto_arima_f` that leads to wrong model frequently being returned. Closes https://github.com/Nixtla/statsforecast/issues/897


jmoralez (approved):
Thanks!

PR #907, opened 2024-09-13
static build requirements

dimbleby:
so that metadata is the same in all distributions

perhaps a pipeline is about to show me why you didn't do it like this all along, but it seems strange...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,9 +49,8 @@ all_requirements = [
     *plotly_requirements,
     *polars_requirements,
     *dev_requirements,
+    *ray_requirements,
 ]
-if sys.version_info < (3, 12):
-    all_requirements.extend(ray_requirements)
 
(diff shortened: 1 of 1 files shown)


dimbleby:
I will not be signing the CLA.

You are welcome to this contribution, take it or leave it.

jmoralez:
Ray doesn't support python 3.12 (https://github.com/ray-project/ray/issues/40211)

dimbleby:
yes it does eg see the wheels at https://pypi.org/project/ray/#files

dimbleby:
but either way it is still desirable to have static build requirements so if that was your intention it would be be better to go something like
```python
ray_requirements = [req + " ; python_version < '3.12'" for req in ray_requirements]
all_requirements.extend(ray_requirements)
```

PR #908, opened 2024-09-13
static build requirements

dimbleby:
so that metadata is the same in all distributions


jmoralez:
Hey. If you're not going to sign the CLA please open an issue instead

Issue #896, opened 2024-09-04
[AutoARIMA] Incorrect default parameters in AutoARIMA 


robert-robison:
### What happened + What you expected to happen

The default values for `allowmean` and `allowdrift` are both `False` for the implementation of `AutoARIMA` in `statsforecast.models`. This is despite:

1. The docstring claiming the defaults are `True`
2. The defaults being `True` in the `statsforecast.arima` implementation.

I understand there is an effort to align the two implementations (referenced here: https://github.com/sktime/sktime/issues/4598#issuecomment-1561908565), but perhaps this small fix should be make quicker, as it can lead to different forecasts in many common cases.

### Issue Severity

Low: It annoys or frustrates me.

PR #918, opened 2024-09-19
breaking: allowmean and allowdrift True by default in AutoARIMA 

V-nguard:
Hello, 

I encountered the same issue as  https://github.com/Nixtla/statsforecast/issues/896, this should fix it.
I couldn't commit without adding the other files; I don't think I missed a step in the contributing guidelines either.


jmoralez:
Hey. Thanks for your contribution. Can you please add the pin `nbdev==2.3.25` to the `settings.ini` file and restore the other files? Seems like that was added in v2.3.28

V-nguard:
Hi, I'm not fully fluent in git, so is there another way to restore the files without creating another commit? Not sure what the intended operations are (like I'm assuming I don't commit the settings.ini change either because that should be another pull request)

jmoralez:
I think you can run `git checkout main -- python` to restore all the python files and then `nbdev_export` again.

It's ok to commit the `settings.ini` file here, it only changes a dev requirement.

V-nguard:
Thanks for the git command. This should work.

jmoralez:
@V-nguard seems like this improves the performance on the M3 dataset. Can you please set this line https://github.com/Nixtla/statsforecast/blob/8e16c0219e407e5152f10d7e51c7322f2c4cc934/experiments/m3/src/evaluation.py#L53 to `4.46` ([logs](https://github.com/Nixtla/statsforecast/actions/runs/10950455619/job/30405783908?pr=918#step:5:50))?

V-nguard:
I did the change, but I can't commit due to ruff. Something about imported numpy but unused. I ran `ruff check --fix *`, and now it's saying
```
experiments/m3/src/evaluation.py:41: error: No overload variant of "__getitem__" of "list" matches argument type "list[str]"  [call-overload]
experiments/m3/src/evaluation.py:41: note: Possible overload variants:
experiments/m3/src/evaluation.py:41: note:     def __getitem__(self, SupportsIndex, /) -> Any
experiments/m3/src/evaluation.py:41: note:     def __getitem__(self, slice, /) -> list[Any]
experiments/m3/src/evaluation.py:42: error: No overload variant of "__getitem__" of "list" matches argument type "str"  [call-overload]
experiments/m3/src/evaluation.py:42: note: Possible overload variants:
experiments/m3/src/evaluation.py:42: note:     def __getitem__(self, SupportsIndex, /) -> Any
experiments/m3/src/evaluation.py:42: note:     def __getitem__(self, slice, /) -> list[Any]
experiments/m3/src/evaluation.py:42: error: No overload variant of "__setitem__" of "list" matches argument types "str", "Any"  [call-overload]
experiments/m3/src/evaluation.py:42: note:     def __setitem__(self, SupportsIndex, Any, /) -> None
experiments/m3/src/evaluation.py:42: note:     def __setitem__(self, slice, Iterable[Any], /) -> None
experiments/m3/src/evaluation.py:43: error: "list[Any]" has no attribute "set_index"  [attr-defined]
experiments/m3/src/evaluation.py:44: error: "list[Any]" has no attribute "columns"  [attr-defined]
experiments/m3/src/evaluation.py:45: error: "list[Any]" has no attribute "set_index"  [attr-defined]
experiments/m3/src/evaluation.py:46: error: "list[Any]" has no attribute "droplevel"  [attr-defined]
experiments/m3/src/evaluation.py:47: error: "list[Any]" has no attribute "to_csv"  [attr-defined]
experiments/m3/src/evaluation.py:48: error: "list[Any]" has no attribute "query"  [attr-defined]
experiments/m3/src/evaluation.py:49: error: "list[Any]" has no attribute "query"  [attr-defined]
experiments/m3/src/evaluation.py:60: error: "dict[str, float]" has no attribute "index"  [attr-defined]
Found 11 errors in 1 file (checked 3 source files)
```

jmoralez:
Yeah, sorry that's mypy. We don't really care about running it on the experiments folder, so can you change this https://github.com/Nixtla/statsforecast/blob/8e16c0219e407e5152f10d7e51c7322f2c4cc934/.pre-commit-config.yaml#L28
to `(experiments|setup.py)`?

V-nguard:
Done!

jmoralez:
Hmm, it seems that the smape for the statistical ensemble in the m3 is increasing in the test, however I'm not able to reproduce that score locally, I've already tried on two different machines and I get what we currently have. I'll keep investigating.

jmoralez:
Oh it was due to me having the `NIXTLA_ID_AS_COL` env variable set, which considered the index when creating the ensemble. I'm able to reproduce the score on the test now, can you please update that one as well? i.e. set this: https://github.com/Nixtla/statsforecast/blob/8e16c0219e407e5152f10d7e51c7322f2c4cc934/experiments/m3/src/evaluation.py#L57 to `4.3`.

jmoralez (approved):
Thanks a lot!

AzulGarza (approved):
awesome @V-nguard and @jmoralez! lgtm

PR #925, opened 2024-10-02
Update broken yahoo query

paullabonne:
Hello, 

I wanted to replicate the GARCH and ARCH doc files but the yahoo query is broken. I have updated it. 

It is my first time contributing so thanks a lot for your contributing instructions; very helpful :) 

Beside the yahoo query there are other bugs preventing the notebooks from running (problems when merging dataset with the unique_id variable); but I thought it is best practice to keep the pull request very specific. I'm happy to help with the other bug if you want.

### DIFF
--- a/settings.ini
+++ b/settings.ini
@@ -21,7 +21,7 @@ ray_requirements = fugue[ray]>=0.8.1 protobuf>=3.15.3,<4.0.0 numpy<2 pandas<2.2 
 dask_requirements = fugue[dask]>=0.8.1
 spark_requirements = fugue[spark]>=0.8.1
 plotly_requirements = plotly plotly-resampler
-dev_requirements = black datasetsforecast fire nbdev==2.3.25 nbformat nbdev_plotly pandas[plot] pmdarima polars[numpy] pre-commit prophet pyarrow pybind11 pytest scikit-learn setuptools<70 supersmoother
+dev_requirements = black datasetsforecast fire nbdev==2.3.25 nbformat nbdev_plotly pandas[plot] pmdarima polars[numpy] pre-commit prophet pyarrow pybind11 pytest scikit-learn setuptools<70 supersmoother yfinance
 nbs_path = nbs
 doc_path = _docs
 recursive = True


nbs/docs/models/GARCH.ipynb
```diff
@@ -233,6 +233,26 @@
     "rcParams['figure.figsize'] = (18,7)"
```
jmoralez, line 233: <p>Please remove this cell</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Nixtla/statsforecast/pull/925/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

jmoralez:
Thank you for your contribution! Would the maintenance be easier if we used something like [yfinance](https://github.com/ranaroussi/yfinance)?

paullabonne:
Yes using yfinance would most probably prevent the query from failing next time yahoo change their API, at the cost of an extra dependency.

jmoralez:
Are you interested in contributing that in this PR? It'd require adding that dependency here https://github.com/Nixtla/statsforecast/blob/9316bc7c875d700692538edd09ad5f6bd7a0d222/settings.ini#L24 (please note that they're sorted alphabetically) and using it in those notebooks.

paullabonne:
I wanted to raise an additional issue for discussion. The cross-validation exercise might not be the most adequate for these models. GARCH and ARCH models are generally used to predict volatility, not so much the raw returns. Since the conditional mean is typically constant, it seems odd to take a draw as the point prediction with `fitted[k] = error*np.sqrt(sigma2[k])`. Another issue with this approach is that simply changing the seed number would probably change the cross-validation result.

It seems more common in the literature to compare the conditional variance with a measure of volatility, like the squared return, see [this paper](https://onlinelibrary.wiley.com/doi/full/10.1002/jae.800) and [this one](https://onlinelibrary.wiley.com/doi/abs/10.1002/1099-131X%28200103%2920%3A2%3C87%3A%3AAID-FOR782%3E3.0.CO%3B2-7) for instance.

Let me know if you think changing the evaluation strategy makes sense. Happy to discuss further!

jmoralez:
Hey. I agree on changing the evaluation strategy, please go ahead.

paullabonne:
ok cool I'll do that in a new PR because it is bit more work

jmoralez (approved):
Thanks!

Issue #903, opened 2024-09-12
seasonal naive confidence intervals are incorrect


andrewscottm:
### What happened + What you expected to happen

The seasonal naive confidence interval for the last time in the forecasted horizon is used for all times.

https://github.com/Nixtla/statsforecast/blob/12f6654dcf6e776af0f3cc461072e8cad6e2f684/python/statsforecast/models.py#L3305

Should be `k = np.floor(np.arange(h) / self.season_length)` as in the [R version](https://github.com/robjhyndman/forecast/blob/0583d2fef45336ee59ef486e24b356520350da9a/R/naive.R#L77).

When the time horizon increases, the first forecasted time's confidence interval increases. It should be constant.

### Versions / Dependencies

statsforecast 1.7.6
pandas 2.2.2
Python 3.11.9 (main, Apr 19 2024, 11:44:45) [Clang 14.0.6 ]
MacOS 14.6.1

### Reproducible example

```python
import pandas as pd
from statsforecast import StatsForecast
from statsforecast.models import SeasonalNaive

df = pd.DataFrame({
    'unique_id': '0',
    'ds': pd.date_range(start='2000-01-01', periods=10, freq='1D'),
    'y': [1., 4., 5., 5., 1., 4., 5., 5., 7., 3.]
})

m = StatsForecast(models=[SeasonalNaive(season_length=7)], freq='1D', n_jobs=1, verbose=True)
m.fit(df=df)
print(m.predict(h=1, level=[95]).iloc[0, :])
print(m.predict(h=8, level=[95]).iloc[0, :])
print(m.predict(h=15, level=[95]).iloc[0, :])
```
unique_id                                0
ds                     **2000-01-11 00:00:00**
SeasonalNaive                          5.0
SeasonalNaive-lo-95              **-1.093776**
SeasonalNaive-hi-95              **11.093776**

unique_id                                0
ds                     **2000-01-11 00:00:00**
SeasonalNaive                          5.0
SeasonalNaive-lo-95                **-3.6179**
SeasonalNaive-hi-95                **13.6179**

unique_id                                0
ds                     **2000-01-11 00:00:00**
SeasonalNaive                          5.0
SeasonalNaive-lo-95              **-5.554729**
SeasonalNaive-hi-95              **15.554729**

### Issue Severity

Medium: It is a significant difficulty but I can work around it.

PR #932, opened 2024-10-23
fix: seasonal naive confidence interval bug

andrewscottm:
Corrects the seasonal naive confidence interval when `h > season_length`, as raised in https://github.com/Nixtla/statsforecast/issues/903

Adds a new test that will fail without this correction. Previously, when `h` was increased from `season_length` to `season_length + 1` the prediction intervals for times `< season_length + 1` would widen. Now they remain unchanged.


andrewscottm:
Hi @jmoralez. Are you able to review this? Many thanks

jmoralez (approved):
Thanks a lot!

MMenchero:
Hi @andrewscottm! can you share with us your email?  Given your contribution to StatsForecast, we would like to connect with you directly. Thanks and regards.

PR #946, opened 2024-11-25
breaking: remove deprecated behavior

jmoralez:
Removes the following behavior, which has been deprecated for about a year:

* Providing `df` and  `sort_df` to the `StatsForecast` constructor. `df` has to be provided to the methods and `sort_df` isn't used anymore, we always check if it's sorted.
* Providing the id as the dataframe index, it has to be a column. Also the outputs from the methods will now always have the id as a column.
* The `ETS` model, in favor of `AutoETS`
* The `seasonal_periods` argument of  `TBATS` and `AutoTBATS`, in favor of `season_length`.

Also updates the documentation to remove remaining usages of this behavior.


AzulGarza (approved):
awesome work, @jmoralez!

PR #966, opened 2025-01-04
fix: Mh update in _class3models

omriKramer:
`Mh` was outside the loop looping over `h`, therefore it didn't update and all variance terms were scaled by the first prediction.


omriKramer:
Anything I can do to help this PR go through review?

PR #973, opened 2025-01-27
enh(SES): improve implementation

christophertitchen:
**Summary**

- Amend `README.md` to note support for in-sample fitted values for ``SimpleExponentialSmoothing``, ``SimpleExponentialSmoothingOptimized``, ``SeasonalExponentialSmoothing``, and ``SeasonalExponentialSmoothingOptimized``.
- ~~Set the first in-sample fitted value in ``_ses_fcst_mse`` to the first observation, rather than `np.nan`, to make it explicitly clear that the initial state is the first observation.~~
- Improve the efficiency and performance of ``_ses_fcst_mse`` by, for example, pre-calculating the smoothing complement, $1 - \alpha$, and removing the unnecessary $n - 1$ copies of fitted values before assignment.
- Improve the performance of ``_optimized_ses_forecast`` by changing from ``scipy.optimize.minimize`` to ``scipy.optimize.minimize_scalar``. This uses Brent's method which is derivative-free and optimised for local one-dimensional optimisation problems like this, with guaranteed convergence within a reasonable number of evaluations. The convergence will also be superlinear for our strictly convex $C^2$ objective function when the minimum is at an interior point of the interval for $\alpha$, which will be the case particularly in ``SimpleExponentialSmoothingOptimized`` and ``SeasonalExponentialSmoothingOptimized``, where $0 \lt \alpha \lt 1$. It should also be quite a bit faster even for the $0.1 \lt \alpha \lt 0.3$ bounds of ADIDA, Croston, and IMAPA. It also does not have the initialisation sensitivity that L-BFGS-B has (the initialisation of $\alpha$ was also out-of-bounds before for some reason).
- ~~_Add a temporary benchmark to `test_efficiency.py` for CodSpeed._~~
- Improve the efficiency and performance of the objective function and forecast function.

**Benchmarks**

**Memory Profile**

*Base*

*Head*


jmoralez:
Thanks. Can you just elaborate on the following?

> Set the first in-sample fitted value in _ses_fcst_mse to the first observation, rather than np.nan, to make it explicitly clear that the initial state is the first observation.

The fitted values are supposed to be the one-step ahead forecasts with an expanding training set, so the first value should be NaN.

christophertitchen:
> Thanks. Can you just elaborate on the following?
> 
> > Set the first in-sample fitted value in _ses_fcst_mse to the first observation, rather than np.nan, to make it explicitly clear that the initial state is the first observation.
> 
> The fitted values are supposed to be the one-step ahead forecasts with an expanding training set, so the first value should be NaN.

Hi José. I reverted the change to the initial fitted value as requested to align with the one-step forecast with an expanding training set definition and made a few more performance improvements.

I originally changed it as I thought that as our implementation uses $x_{0}$ as $\ell_{0}$ to initialise the smoothing process, it should be the initial fitted value. It works this way in other libraries, like the example from _tsintermittent_ below which uses the same approach as us to set $\ell_{0}$, and of course also as expected in the SSOE state space approaches in _forecast_ and _smooth_.

python/statsforecast/models.py
```diff
@@ -2129,38 +2129,62 @@ def __init__(
         )
 
 # %% ../../nbs/src/core/models.ipynb 129
-@njit(nogil=NOGIL, cache=CACHE)
-def _ses_fcst_mse(x: np.ndarray, alpha: float) -> Tuple[float, float, np.ndarray]:
-    r"""Perform simple exponential smoothing on a series.
+@njit(nogil=NOGIL, cache=CACHE, fastmath=True)
```
jmoralez, line 2132: Please remove the fastmath, we support float32 and that can loose a lot of precision with this.

python/statsforecast/models.py
```diff
@@ -2129,38 +2129,62 @@ def __init__(
...
+def _ses_sse(alpha: float, x: np.ndarray) -> float:
```
jmoralez, line 2133: Is this really worth it? Both of the functions you extracted are basically the same, this just avoids filling the fitted values but I doubt that makes a huge difference.

python/statsforecast/models.py
```diff
@@ -2129,38 +2129,62 @@ def __init__(
...
+    Returns
+    -------
+    sse : float
+        Residual sum of squares for the fit.
 
-    mse /= n
-    forecast = alpha * x[-1] + (1 - alpha) * smoothed
-    return forecast, mse, fitted
+    """
+    complement = 1 - alpha
```
jmoralez, line 2149: This will most likely be optimized by the compiler, so I don't think we need this indirection, we're talking 1 microsecond here, which is the improvement shown in the plot and I guess comes from not initializing the fitted array values.

jmoralez (changes requested):
Thanks but I think the big win here is just using brent and we can avoid complicating things.

python/statsforecast/models.py
christophertitchen: Is there an issue with manually hoisting $1 - \alpha$ out of the loop though, rather than expecting LLVM to recognise that it's loop invariant?

I just had a quick look at the IR and for some reason it doesn't seem to hoist it out of the loop without the indirection, which actually seems to be the case if you look at the two benchmarks below for a long time series.

image

Are you happy with me leaving it as is?

python/statsforecast/models.py
christophertitchen: I don't think we actually lose a lot of precision in this case, and the performance improvement from enabling the associative property and floating-point contraction is nice.

The code below compares the results with and without the fastmath flags enabled, and doesn't raise an exception even at very high accuracy, 1e-12.

image

Do you still want me to remove it?

jmoralez: Yes, this will end up being migrated to C++ and we won't use the fast-math flag, so we will get different results with this.

python/statsforecast/models.py
jmoralez: Yes, pelase.

python/statsforecast/models.py
jmoralez: I think we can keep this function and change the other one.

python/statsforecast/models.py
```diff
@@ -2129,38 +2129,62 @@ def __init__(
...
-    _, mse, _ = _ses_fcst_mse(x, alpha)
-    return mse
+    return sse
 
 
+@njit(nogil=NOGIL, cache=CACHE, fastmath=True)
 def _ses_forecast(x: np.ndarray, alpha: float) -> Tuple[float, np.ndarray]:
-    r"""One step ahead forecast with simple exponential smoothing."""
-    forecast, _, fitted = _ses_fcst_mse(x, alpha)
+    r"""Compute the one-step ahead forecast for a simple exponential smoothing fit.
```
jmoralez, line 2162: To avoid duplicating code and having to compile two functions we can use coreforecast instead:

```python
from coreforecast.exponentially_weighted import exponentially_weighted_mean

def _ses_forecast(x: np.ndarray, alpha: float) -> Tuple[float, np.ndarray]:
    fitted = np.append(np.nan, exponentially_weighted_mean(x[:-1], alpha))
    forecast = alpha * x[-1] + (1 - alpha) * fitted[-1]
    return forecast, fitted
```

jmoralez (changes requested):

python/statsforecast/models.py
christophertitchen: I didn't know that was in coreforecast! I made a slight change to use `exponentially_weighted_mean` for the forecast too, so it means copying the last element (one-step ahead forecast) to a Python float, and avoiding creating a copy of the array using `np.append` by instead overwriting `fitted` in-place to return the correct fitted values to improve performance. 👍 

```python
def _ses_forecast(x: np.ndarray, alpha: float) -> Tuple[float, np.ndarray]:
    r"""Compute the one-step ahead forecast for a simple exponential smoothing fit.

    Parameters
    ----------
    x : numpy.array
        Clean time series of shape (n, ).
    alpha : float
        Smoothing parameter.

    Returns
    -------
    tuple of (float, numpy.array)
        One-step ahead forecast and in-sample fitted values.

    """
    fitted = exponentially_weighted_mean(x, alpha)
    forecast = fitted.item(-1)
    fitted[1:] = fitted[:-1]
    fitted[0] = np.nan
    return forecast, fitted
```

python/statsforecast/models.py
christophertitchen: Got it.

christophertitchen:
@jmoralez tests are failing, see my PR nixtla/coreforecast#79 for fix to C++ code for EMA, which should fix the issue. 🤞

jmoralez:
@christophertitchen can you restore your numba function? It'll take me a bit longer to do the coreforecast release because I want to refactor all the single array functions. Sorry for the troubles.

christophertitchen:
> @christophertitchen can you restore your numba function? It'll take me a bit longer to do the coreforecast release because I want to refactor all the single array functions. Sorry for the troubles.

No worries!

jmoralez (approved):
Thanks!

Issue #976, opened 2025-02-04
[<Library component: Model|Core|etc...>]


moagstar:
### What happened + What you expected to happen

It is not possible to run statsforecast when running with `PYTHONOPTMIZE=2` since there is some automatic formatting being applied to __doc__, for example:

```python
__init__.__doc__ = __init__.__doc__.format(**_param_descriptions)
```

### Versions / Dependencies

<details><summary>Click to expand</summary>
Dependencies:

</details>

### Reproducible example

```python
__init__.__doc__ = __init__.__doc__.format(**_param_descriptions)
```

### Issue Severity

None

robrui:
The bug is actually in `coreforecast` (a dependency of statsforecast). The docstring decorators in `expanding.py` and `rolling.py` called `.format()` on `__doc__` without checking for `None`.

Fixed in Nixtla/coreforecast#131 — added `if function.__doc__:` guards before formatting.

Once that fix is merged and released, statsforecast will work with PYTHONOPTIMIZE=2 automatically.

jbbqqf:
Hi! Triaging older issues this week — I think this one has shipped and can be closed.

I tried to reproduce the original symptom against current `main`:

```bash
PYTHONOPTIMIZE=2 python -c "
import statsforecast
from statsforecast import StatsForecast
from statsforecast.models import Naive, AutoARIMA, AutoTheta, AutoETS, AutoCES, MSTL, MFLES
print('imports ok')
"
# → imports ok
```

`grep -rn '__doc__' python/statsforecast` finds no `__init__.__doc__ = __init__.__doc__.format(...)` patterns anymore — the codebase moved from `lazydocs` / nbdev-style notebook-generated docstrings to `mkdocstrings` ([#1060](https://github.com/Nixtla/statsforecast/pull/1060)), so the line you flagged no longer exists.

Would you mind closing this out? If you can still trigger an `AttributeError` under `PYTHONOPTIMIZE=2` against `main`, point me at the exact failing module/symbol and I'll dig further.

(Disclosure: I drafted this comment with help from Claude Code while triaging stale issues; the references above were verified manually against the current `main` branch.)

PR #977, opened 2025-02-04
fix: ensure statsforecast can run with PYTHONOPTIMIZE=2

moagstar:
Fix for https://github.com/Nixtla/statsforecast/issues/976


jmoralez:
Thanks! Can you please export the notebooks (run `nbdev_export`)?

Alternatively you can just install the pre-commit hooks as described in our [contributing guide](https://github.com/Nixtla/statsforecast/blob/main/CONTRIBUTING.md) and that'll export them.

moagstar:
@jmoralez ah yes sorry I missed that step - I had some trouble with the pre-commit hooks, but I've managed to generate the files manually now.

moagstar:
@jmoralez I see that some checks are failing.

nbdev_export.............................................................Failed
- hook id: nbdev_export
- files were modified by this hook
Error: Process completed with exit code 1.

When I ran nbdev locally it did generate some python files for notebooks I hadn't touched - but one of those generated files had some mypy errors:

python/statsforecast/distributed/multiprocess.py:35: error: Unexpected keyword argument "df" for "_StatsForecast"  [call-arg]
python/statsforecast/distributed/multiprocess.py:47: error: Unexpected keyword argument "df" for "_StatsForecast"  [call-arg]

So I just added the python files for the two notebooks I had touched.
Do you have some idea what might be happening here?

jmoralez:
We use a specific version of nbdev https://github.com/Nixtla/statsforecast/blob/52a5c9abfff8d594e8a2190add88fb224067bd3b/settings.ini#L24

so you have to set up the environment as described in the contributing guidelines, or I can fix the linting errors if you want.

python/statsforecast/core.py
```diff
@@ -1,3 +1,5 @@
+"""Methods for Fit, Predict, Forecast (fast), Cross Validation and plotting"""
+
```
jmoralez, line 2:
```suggestion
```

jmoralez (changes requested):

python/statsforecast/core.py
jmoralez: This should fix the linter error, I don't have permission to push to your branch.

Issue #1006, opened 2025-04-24
Memory crash/performance issue within arima.py when using exogenous regressors


belliriccardo:
### What happened + What you expected to happen

So I am trying to use AutoARIMA for time series forecasting. When I use the model as an univariate model, everything is fine. But, when I try to use it with exogenous regressors, the issues begin: python (conda) starts using a lot more memory, the computer starts swapping like crazy and I consistently get BSOD'ed, lol (I am on windows).
Having debugged the code, I found out the places where this happens: it's on both usages of `numpy.linalg.svd`; this methods inherently uses a lot of memory for large matrices, and since my time series is quite long this causes the issue. The points of failure are these two:
[https://github.com/Nixtla/statsforecast/blob/1342c1e69a54172784d71ce10b633a96cefb2934/python/statsforecast/arima.py#L404](https://github.com/Nixtla/statsforecast/blob/1342c1e69a54172784d71ce10b633a96cefb2934/python/statsforecast/arima.py#L404) and [https://github.com/Nixtla/statsforecast/blob/1342c1e69a54172784d71ce10b633a96cefb2934/python/statsforecast/arima.py#L1454](https://github.com/Nixtla/statsforecast/blob/1342c1e69a54172784d71ce10b633a96cefb2934/python/statsforecast/arima.py#L1454)
For example, in the second case, my matrix `X` has shape (30000, 2). I think a solution exists and might also be quite simple: since it is so rectangular, in the first case, for `xreg`, the parameter `full_matrices=False` can be used, and in the second since we are only interested in the singular values we could just use the parameter `compute_uv=False`. Even for time series that are not as large as mine, using these options greatly speed up the calculation, especially in the second case (by not calculating U and V), so everyone could benefit from such a change.
I also found out that using `scipy`'s SVD method uses less memory and manages not to crash my computer (though it hogs it quite a bit as well).
Also a disclaimer: I am not, by any means, an expert in this field! If these changes invalidate some of the logic of the implementation/model, different approaches should be taken!

### Versions / Dependencies

<details><summary>Click to expand</summary>
Dependencies:

- Latest statsforecast 2.0.1

- conda 3.12.9
</details>

### Reproducible example

Here's my proposed workaround; regarding the first [usage](https://github.com/Nixtla/statsforecast/blob/1342c1e69a54172784d71ce10b633a96cefb2934/python/statsforecast/arima.py#L403):
```python
if not orig_xreg:
[...]

PR #1015, opened 2025-05-30
Fix memory crash/perf. issue with arima.py + exogenous regressors

belliriccardo:
This simple patch would fix the issue described in #1006; also, by not computing the full matrices in one case and just the singular values in the other, there would also be big performance gains, depending on the time series, regressors, etc.


elephaint:
@belliriccardo Thanks for your contribution!

elephaint (changes requested):
Seems like a simple improvement, thanks!

Note that we use nbdev to create the code; hence, please run `nbdev_export` when making changes to the ipynb files.

belliriccardo:
Hey @elephaint ! Sorry I'm late, I've just seen this. Thanks for the feedback; I didn't know about `nbdev_export`, but I can see that you already [merged the changes](https://github.com/Nixtla/statsforecast/pull/1015/commits/5c3527df8c06659f3c75059b8bc183e4d2235356). Do I still need to run the command or do something else?

Issue #1018, opened 2025-07-04
import error with latest scipy 1.16.0


elephaint:
### What happened + What you expected to happen

See here: https://github.com/statsmodels/statsmodels/issues/9584

This issue breaks basically every test in all our repos on Python>3.10.

Solution is to pin statsmodels until they release a new version.

### Versions / Dependencies

2.0.1

### Reproducible example

```python
# paste your code here
```

### Issue Severity

None

PR #1019, opened 2025-07-04
[FIX] Pin statsmodels / scipy

elephaint:
- Temporary fix for #1018
- Update deps according to #1016 to ensure tests complete.


elephaint:
It appears the errors with DuckDB are resolved when upgrading the manylinux image as in #1016. If that one is merged, I think this one can be closed.

PR #1030, opened 2025-07-14
rm `df` arg from constructor

deven367:
`df` is an arg of `forecast` and `cross_validation`, due to this I had some failing testcases with `pytest`

image

cc @MMenchero @elephaint


elephaint (dismissed):
LGTM. Good to merge once linting is fixed.

elephaint (approved):
Thanks, lgtm

PR #1055, opened 2025-10-12
fix: update broken model documentation links in README

colesussmeier:
Currently all of the links to model documentation in the README.md are broken

I've updated these to point to their respective model references, with a few exceptions:

AutoMFLES, AutoTBATS, TBATS, and all baseline models do not have their own model reference sections, so they are linked directly to API references instead

I also changed the spelling of intermittent in a section header since it was right next to these other tweaks


nasaul:
Thank you so much for taking the time to contribute to StatsForecast! We really appreciate your effort and initiative in improving the documentation.

However, we're currently working on a new documentation pipeline, and we need to complete that infrastructure before we can accept documentation PRs. This will ensure that all documentation contributions align with our new system and standards.

We'll keep your PR in mind once the new pipeline is ready. Thanks again for your understanding and for wanting to contribute to the project!

PR #1058, opened 2025-10-15
chore: exclude *.md from ci.yml

mergenthaler:


deven367:
The `paths-ignore` arg is slightly tricky, if we modify any file other than `*.md` like the parser file or `build-docs.yml` or even the Makefile, all tests will run

PR #1060, opened 2025-10-20
[docs] lazydocs to mkdocstrings

deven367:
This PR migrates from `lazydocs` to `mkdocstrings`

- [x] update Makefile
- [x] update workflow
- [x] add docs
- [x] add parser


deven367:
@nasaul `_param_descriptions` that we removed also needs to refactored in `statsforecast.distributed.fugue`

Issue #1048, opened 2025-09-10
Unpin scipy


melanopsis:
SciPy was pinned via #1019 as a temporary fix to deal with statsmodels' SciPy [issue](https://github.com/statsmodels/statsmodels/issues/9584). The recent statsmodels [v0.4.15](https://github.com/statsmodels/statsmodels/releases/tag/v0.14.5) seems to have taken care of this. Please remove the `scipy < 1.16` requirement accordingly. 

Thanks!

PR #1074, opened 2026-01-06
Unpin scipy

nasaul:
Unpins scipy and update statsmodels accordingly.
Solves #1048 

There were some issues while doing the tests for this PR which aren't particular to the unpining of `scipy`:

- Pyspark pinning version in Windows. There's an issue that it's already been solved in [here](https://github.com/apache/spark/pull/53546) however there isn't a new release with the fix yet.
- Removes Ray testing in Windows since it's still in beta in [Ray 2.53.0](https://docs.ray.io/en/latest/ray-overview/installation.html)
- Scipy stills needs pin down for Python 3.9 since they dropped support in [here](https://github.com/scipy/scipy/pull/20373)
- Updates fugue version to support Ray latest versions
- Installs same in Ray as in all for testing
- Adds Ray session config for testing (without this MacOS with Python 3.9 always hangs)


elephaint (approved):
LGTM

PR #1079, opened 2026-01-21
fix: validate unique model names for models without alias

HummerQAQ:
## Issue

This PR fixes a gap in model name validation when using custom models without an explicit `alias`.

While working on a research project, I defined two simple custom forecasting models (`modelA` and `modelB`) that do not expose an `alias` attribute. During experimentation, I accidentally passed duplicated models:

```python
models = [modelA(), modelA()]
```

However, StatsForecast did not raise an error and instead silently merged both outputs into a single column in the forecast DataFrame. This made it appear as if one model was missing, leading to unnecessary debugging time before identifying the root cause.

## What changed
1. Model name validation now:
- Uses alias when available
- Falls back to repr(model) when alias is not defined
- Enforces uniqueness across all models during initialization
2. A clear `ValueError` is raised during initialization if duplicate model names are detected
3. Added regression tests to ensure duplicate names are rejected even when models do not expose alias

## Why this matters
While most users rely on built-in models, StatsForecast’s API allows users to pass arbitrary custom models. In such advanced or experimental scenarios, silent configuration errors might be costly to debug.


nasaul (approved):
Thanks for your contribuiton!

PR #1072, opened 2026-01-02
[FEAT] Simulate method for sample trajectories

ankitlade12:
## Description
This PR significantly expands the [simulate](cci:1://file:///Users/ankithemantlade/Desktop/nixtla_statsforecast/python/statsforecast/core.py:465:4-505:68) method capabilities across the [StatsForecast](cci:2://file:///Users/ankithemantlade/Desktop/nixtla_statsforecast/python/statsforecast/core.py:1704:0-1957:9) library. Beyond providing basic sample trajectories, it now supports a wide range of error distributions, enabling high-fidelity probabilistic forecasting and robust scenario analysis. 

This implementation centralizes simulation logic, improves model accuracy by incorporating historical residuals, and ensures high consistency across the model library.

## Key Changes
- **Advanced Error Distributions**:
    - Introduced a centralized [statsforecast/simulation.py](cci:7://file:///Users/ankithemantlade/Desktop/nixtla_statsforecast/python/statsforecast/simulation.py:0:0-0:0) utility for sampling from multiple distributions: `normal`, `t` (Student's t), `bootstrap` (empirical), `laplace`, `skew-normal`, and `ged` (Generalized Error Distribution).
    - Implemented calibrated variance scaling for all distributions to ensure generated noise matches the model's estimated [sigma](cci:1://file:///Users/ankithemantlade/Desktop/nixtla_statsforecast/python/statsforecast/tbats.py:1014:0-1030:17).
- **Expanded Model Coverage**:
    - **Smoothing Models**: Added [simulate](cci:1://file:///Users/ankithemantlade/Desktop/nixtla_statsforecast/python/statsforecast/core.py:465:4-505:68) methods and analytic prediction intervals to [SimpleExponentialSmoothing](cci:2://file:///Users/ankithemantlade/Desktop/nixtla_statsforecast/python/statsforecast/models.py:2243:0-2426:18) (Standard & Optimized) and [SeasonalExponentialSmoothing](cci:2://file:///Users/ankithemantlade/Desktop/nixtla_statsforecast/python/statsforecast/models.py:2650:0-2851:18) (Standard & Optimized).
[...]

PR #1083, opened 2026-01-28
Add generate() method for sample forecast trajectories

### LINKED ISSUE #1023: Support for generating sample trajectories
### Description

Currently, StatsForecast models provide point forecasts and prediction intervals, but lack a direct method to generate predictive samples from the fitted models.
A built-in method to generate predictive samples from fitted models such as [generate](https://fable.tidyverts.org/reference/generate.ARIMA.html) would make StatsForecast more competitive with R's fable ecosystem. This would also allow us to switch from fable to nixtla in our [workflow](https://github.com/CDCgov/pyrenew-hew/blob/main/pipelines/forecast_timeseries_ensemble.R) which relies on using sample trajectories.

### Use case

A method to generate predictive samples/full predictive distribution from fitted models.

IndarKarhana:
## Description

Closes #1023

This PR adds a `generate()` method to `AutoARIMA`, `AutoETS`, and `AutoCES` models for generating sample forecast trajectories, similar to R's fable [`generate.ARIMA`](https://fable.tidyverts.org/reference/generate.ARIMA.html).

### Motivation

The CDC ([pyrenew-hew](https://github.com/CDCgov/pyrenew-hew)) requested this feature for epidemiological modeling workflows that require Monte Carlo simulation with correlated forecast paths. The existing `predict(level=[...])` method only returns quantile bounds, which loses temporal correlation information needed for downstream simulation.

### What's New

**API Addition:**
```python
# Generate 500 sample forecast trajectories
samples = model.generate(h=24, n_samples=500, bootstrap=False, random_state=42)
# Returns: np.ndarray of shape (500, 24)
```

**Comparison with existing approach:**

| Feature | `predict(level=[...])` | `generate(n_samples=1000)` |
|---------|----------------------|---------------------------|
| Output | Quantile bounds only | Full sample paths (1000 × h) |
| Get quantiles | ✅ Any level | ✅ Compute from samples |
| Raw sample paths | ❌ Not available | ✅ Full access |
| Monte Carlo simulation | ❌ No paths to simulate | ✅ Feed paths to downstream models |
| Temporal correlation | ❌ Each quantile computed independently | ✅ Each path is temporally coherent |

The key difference: `predict()` gives you summary statistics, while `generate()` gives you the underlying samples that those statistics come from. This is essential for Monte Carlo workflows where you need to propagate correlated forecast uncertainty through downstream calculations.

### Implementation Approach

The existing codebase has some simulation-related scaffolding (e.g., `bootstrap` parameters, `npaths` arguments) but the full sample trajectory generation wasn't exposed to users. This PR builds on that foundation:

| Model | Approach |
|-------|----------|
| `AutoARIMA` | Kalman state-space simulation with drift handling |
| `AutoETS` | ETS state-space forward simulation |
| `AutoCES` | Extended existing internal simulation to return raw paths |

### Parameters

```python
def generate(
    self,
    h: int,                          # Forecast horizon
    n_samples: int = 100,            # Number of sample trajectories
    bootstrap: bool = False,         # Resample from residuals vs N(0, σ²)
    random_state: Optional[int] = None,  # Reproducibility
) -> np.ndarray:
```

### New Files
- `nbs/docs/tutorials/SampleTrajectories.ipynb` - Tutorial with visual comparison
- `python/statsforecast/tests/test_generate.py` - 30 comprehensive tests

### Modified Files
- `python/statsforecast/models.py` - Added `generate()` to AutoARIMA, AutoETS, AutoCES
- `python/statsforecast/arima.py` - Added `generate_arima_samples()` helper
- `python/statsforecast/ets.py` - Added `generate_ets_samples()` helper
- `python/statsforecast/ces.py` - Added `generate_ces_samples()` helper

## Testing

```bash
pytest python/statsforecast/tests/test_generate.py -v
# 30 tests passed
```

Test coverage includes:
- Basic functionality for all three models
- Seasonal models (season_length=12)
- Reproducibility with `random_state`
- Bootstrap mode
- Error handling (calling before fit)
- Sample mean matching point forecast

### Side-by-side Comparison

The tutorial notebook includes a visual comparison showing the difference between:
- **LEFT**: Existing `predict(level=[80, 95])` - only shaded quantile bounds
- **RIGHT**: New `generate(n_samples=500)` - individual sample trajectories visible

This makes the value proposition immediately clear for Monte Carlo use cases.

### DIFF
--- a/nbs/docs/tutorials/SampleTrajectories.ipynb
+++ b/nbs/docs/tutorials/SampleTrajectories.ipynb
@@ -1,0 +1,229 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#| echo: false\n",
+    "import logging\n",
+    "import warnings\n",
+    "warnings.simplefilter('ignore')\n",
+    "logging.getLogger('statsforecast').setLevel(logging.ERROR)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Generating Sample Forecast Trajectories\n",
+    "\n",
+    "> Learn how to generate sample forecast paths for probabilistic forecasting and simulation"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Introduction\n",
+    "\n",
+    "While prediction intervals provide a useful summary of forecast uncertainty, **sample trajectories** provide a richer representation by generating many possible future scenarios.\n",
+    "\n",
+    "This is especially useful for:\n",
+    "\n",
+    "- **Monte Carlo simulation**: Simulating many possible futures\n",
+    "- **Coherent probabilistic forecasts**: Samples that respect temporal dependencies\n",
+    "- **Risk analysis**: Understanding the range of possible outcomes\n",
+    "\n",
+    "The `generate()` method provides this capability for `AutoARIMA`, `AutoETS`, and `AutoCES` models."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Install Libraries"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# !pip install statsforecast"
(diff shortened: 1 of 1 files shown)


nasaul:
Hello @IndarKarhana thanks for your contribution and we're always grateful for this. However currently a user already submitted a solution in #1072 that covers an extra model (theta) and opens up for more distributions in order to sample the errors from.

PR #1113, opened 2026-03-12
[FEAT] Adds t, ged, normal-skewed and laplace distributions for fitting ARIMA

nasaul:
This PR's adds different likelihoods functions for fitting the ARIMA family models.

- [x] Implemented different likelihood functions
- [x] Adds testing


python/statsforecast/arima.py
```diff
@@ -594,10 +869,29 @@ def arma_css_op(p, x):
...
         val = (val[0], val[3])
-        sigma2 = val[0] / n_used
-
-    value = 2 * n_used * res.fun + n_used + n_used * np.log(2 * np.pi)
-    aic = value + 2 * sum(mask) + 2 if method != "CSS" else np.nan
+        if distribution == "normal":
+            sigma2 = val[0] / n_used
+        elif distribution == "laplace":
+            # sigma2 stores b̂ (the fitted Laplace scale)
+            sigma2 = np.nansum(np.abs(val[1])) / n_used
```
MMenchero: I think what is being saved here is $\hat{b}$, not $\sigma^2$, which for Laplace is $\sigma^2 = 2\hat{b}^2$. 

```python 
b_hat = np.nansum(np.abs(val[1])) / n_used
sigma2 = 2.0 * b_hat**2
```

nasaul:
@elephaint I think that if we proceed with this, then your C++ implementation will need to incorporate this. Any opinions on this?

python/statsforecast/arima.py
```diff
@@ -365,6 +535,15 @@ def maInvert(ma):
         ncxreg += 1
         nmxreg = ["intercept"] + nmxreg
 
+    if distribution not in _VALID_DISTRIBUTIONS:
+        raise ValueError(
+            f"distribution must be one of {_VALID_DISTRIBUTIONS}, got {distribution!r}"
+        )
+    if distribution != "normal" and method == "CSS":
```
JQGoh: nit: We can consider taking this chance to introduce enums/constant variables for options like "normal", "CSS" etc. to improve code readability.

python/statsforecast/utils.py
```diff
@@ -266,15 +268,28 @@ def _naive(
     return fcst
 
 
+_VALID_DISTRIBUTIONS = ("normal", "laplace", "t", "skew-normal", "ged")
```
JQGoh: ditto: replace distribution functions with enums/const variables than strings

python/statsforecast/utils.py
```diff
@@ -266,15 +268,28 @@ def _naive(
...
+    if distribution == "laplace":
+        return laplace_dist.ppf(p)
+    elif distribution == "t":
+        return t_dist.ppf(p, df=dist_params.get("nu", 5.0))
+    elif distribution == "skew-normal":
+        return skewnorm_dist.ppf(p, a=dist_params.get("alpha_dist", 0.0))
+    elif distribution == "ged":
+        return gennorm_dist.ppf(p, beta=dist_params.get("beta_dist", 2.0))
+    else:
+        return norm.ppf(p)
```
JQGoh: nit: A dictionary could be more readable and raise key-error exception when unexpected distribution name is given


```suggestion
    dist_ppf = {
        "laplace": laplace_dist.ppf(p),
        "t": t_dist.ppf(p, df=dist_params.get("nu", 5.0))
        "skew-normal": skewnorm_dist.ppf(p, a=dist_params.get("alpha_dist", 0.0)),
        "ged": gennorm_dist.ppf(p, beta=dist_params.get("beta_dist", 2.0))
        "normal": "norm.ppf(p)
    }
    return dist_ppf[distribution]
```

python/statsforecast/arima.py
JQGoh: More importantly, `distribution` and `method` are input parameters. I suggest that we should fail fast by moving this validation logic soon after the we receive the inputs, placed them around the line 212

python/statsforecast/arima.py
```diff
@@ -809,7 +1114,8 @@ def myarima(
             fit["ic"] = fit[ic]
         else:
             fit["ic"] = fit["aic"] = fit["bic"] = fit["aicc"] = math.inf
-        fit["sigma2"] = np.nansum(fit["residuals"] ** 2) / (nstar - npar + 1)
+        if distribution == "normal":
+            fit["sigma2"] = np.nansum(fit["residuals"] ** 2) / (nstar - npar + 1)
```
JQGoh, line 1115: may I check that is the computation 'sigma2' only relevant for normal distribution? Could this be available for the other distributions?

If sigma2 is not available (same for line 1407 `tmp["sigma2"]`), how does it affect the subsequent steps during the computations?

python/statsforecast/arima.py
```diff
@@ -272,6 +274,174 @@ def armafn(p, x, trans):
             return math.nan
         return 0.5 * (math.log(s2) + res[1] / res[2])
 
+    def armafn_laplace(p, x, trans):
+        x = x.copy()
+        par = coef.copy()
```
JQGoh, line 288: I understand that this follows the style/implementation of `armafn()` function. However, I consider passing the variable `coef` as as an input argument instead such that it can help to communicate better about the intention of the function, rather than assuming `coef` is pre-defined elsewhere of the code.

Same for the other variables like `ncxreg` etc.

JQGoh (commented):
Left some minor comments

JQGoh (approved):

PR #1116, opened 2026-03-19
[FEAT] Unified error distributions across the StatsForecast suite (ETS, CES, Theta + shared C++/Python infra)

nasaul:
## Summary

This PR began as "add non-normal error distributions for fitting ETS" but its scope has expanded into **unified error-distribution infrastructure across the StatsForecast suite**. It introduces a single C++ header and a single Python module that own all distribution math and MLE-fitting helpers, then wires four models — **ARIMA, ETS, AutoCES, and AutoTheta** — through them.

Supported distributions: `normal`, `laplace`, `t`, `skew-normal`, `ged`.

## Motivation

PR #1113 added distribution-aware MLE fitting to ARIMA, and the original version of this PR added it to ETS — independently, duplicating the C++ likelihood math and splitting the Python distribution utilities between `utils.py` and `simulation.py`. This PR consolidates that into shared layers so each future method (and each future distribution) is a small, single-recipe extension rather than another copy of the math.

## What's included

**Shared infrastructure**
- `include/statsforecast/distributions.h` — single C++ source of truth: the `Distribution` enum and per-observation negative-log-likelihood cores (`negloglik_laplace/t/skewnorm/ged`).
- `python/statsforecast/distributions.py` — single Python hub: re-exports the `Distribution` enum / quantile utilities and adds shared fitting helpers (`switch_distribution`, `dist_init_params`, `extract_dist_params`, `aic_bic_aicc`, `error_params_from_model`).

**Model integration**
- **ETS** — refactored to use both shared layers with no behavior change (proving the infra; the existing distribution tests are the regression guard).
- **ARIMA** — distribution imports now point at the shared hub (`distributions.py`); no numeric change.
- **AutoCES** — new distribution-aware one-pass fitting (`optimize_dist` in `ces.cpp`), distribution-aware prediction intervals, and `distribution=` threaded through `cesmodel`/`auto_ces`/`AutoCES`.
- **AutoTheta** — new C++ likelihood-based fitting for non-normal distributions (`optimize_dist` in `theta.cpp`) with AIC-based model selection; normal path unchanged. `distribution=` threaded through `thetamodel`/`auto_theta`/`AutoTheta`.

**Simulation**
- `.simulate()` now defaults `error_distribution` to the fitted `model_["distribution"]` and emits a `UserWarning` on an explicit mismatch (it still proceeds). `SUPPORTED_DISTRIBUTIONS` kept as a deprecated alias.

**Tests**
- `tests/test_distributions.py` — unified parametrized suite covering the model-dict key contract across ARIMA/ETS/CES/Theta, scipy quantile/log-likelihood cross-checks, prediction-interval ordering, and the simulate-mismatch warning.

## Notes / limitations

- Distribution is **user-fixed**: `auto_ces`/`auto_theta` select only the structural type (seasontype / modeltype), not the distribution.
- CES normal `lik` is the concentrated `n·log(SSE)`, while the non-normal objective is a full `-2logL`; these are not comparable across distributions — fine because the distribution is fixed per fit.
- Fitting everywhere is one-pass: Python seeds the extended `x0`, C++ runs a single Nelder-Mead.

## Out of scope (future PRs)

- AutoMFLES, AutoTBATS, GARCH
- Additional distributions beyond the five above

## Checklist

- [x] Shared C++ header + Python hub
- [x] ETS/ARIMA refactored onto shared layers (no behavior change)
- [x] CES & Theta distribution-aware fitting + intervals
- [x] Simulation defaults to fitted distribution + mismatch warning
- [x] Unified parametrized test suite


python/statsforecast/ets.py
```diff
@@ -1198,8 +1367,23 @@ def _compute_pred_intervals(model, forecasts, h, level):
         if math.isnan(phi):
             phi = 0
 
+        dist = model.get("distribution", "normal")
         for k in range(nsim):
-            e = np.random.normal(0, np.sqrt(sigma), h)
+            if dist == "laplace":
```
marcopeix: Can't we do something like this instead of the `elif` ladder?

```python
from statsforecast.distributions import error_params_from_model
e = sample_errors(
    size=h,
    sigma=np.sqrt(sigma),
    distribution=model.get("distribution", "normal"),
    params=error_params_from_model(model),
    residuals=model.get("residuals"),
    rng=...,  
)
```

I think we can also get rid of the scipy imports then.

marcopeix (approved):
Very nice!

PR #1125, opened 2026-04-05
Fix exploding conformal intervals by supporting stable-space transformations

eskayML:
Following our conversation here 

https://github.com/Nixtla/statsforecast/issues/1119

 Please verify if it goes against the original design of the way the module is intended to go

**Problem:**
  When using ConformalIntervals on models trained on transformed data (e.g., log-transformed), the conformity scores are
  calculated in that transformed space. When these intervals are back-transformed to the original scale (e.g., using expm1), the 
  errors also scale exponentially, leading to nonsensical or "exploding" intervals. This was reported in issue #1119.

  **Solution:**
  This PR adds optional transform and back_transform parameters to the ConformalIntervals class. This allows the model to compute  residuals in a stable space (like the original data scale) even if the model itself is operating in a different space (like log  space).

  **Changes:**
   - utils.py: Enhanced ConformalIntervals to accept transform and back_transform callables.
   - models.py: 
       - Updated _conformity_scores to apply the transform before calculating absolute errors.
       - Updated _add_conformal_distribution_intervals to apply the transform to the mean and back_transform to the resulting    
         quantiles.
   - tests/test_models.py: Added test_conformal_intervals_transform to verify that intervals remain stable when using
     np.log1p/np.expm1 transformations.

  **Verification:**
  Added a new test case that reproduces the "exploding" scenario and confirms that providing the transformation functions keeps  
  the intervals within reasonable, stable bounds.


MMenchero:
Thanks for working on this, @eskayML. However, we can’t merge it for several reasons.

1. Copilot raised several high-severity issues, including a signature change that would break existing internal call sites that call `_add_conformal_distribution_intervals` without the new `prediction_intervals` argument, as well as unused imports that would cause CI to fail.
2. It does not fix the default behavior. In fact, the reproducible example from the original issue does not show any explosion I’ve included the code I used and the plot below to show this. 

That said, your contribution is important because it can help guide a proper fix. The issue described in #1119 is real, but whether the transformed prediction intervals explode depends on the magnitude of the conformal scores. This is mainly driven by data variability. 

```python 
from statsforecast.utils import ConformalIntervals
import numpy as np
import pandas as pd

np.random.seed(42)
n = 260  # 5 years weekly
y = np.random.lognormal(mean=6, sigma=0.5, size=n)
df = pd.DataFrame({
    'unique_id': ['series_1'] * n,
    'ds': pd.date_range('2020-01-06', periods=n, freq='W-MON'),
    'y': y
})

# Log transform before fitting
df['y'] = np.log1p(df['y'])

sf = StatsForecast(models=[AutoARIMA(season_length=52, D=1)], freq='W-MON')
ci = ConformalIntervals(h=26, n_windows=3)

fcst = sf.forecast(df=df, h=26, prediction_intervals=ci, level=[70, 90])

# Back-transform
df['y'] = np.expm1(df['y'])
for col in ['AutoARIMA', 'AutoARIMA-lo-70', 'AutoARIMA-hi-70', 'AutoARIMA-lo-90', 'AutoARIMA-hi-90']:
    fcst[col] = np.expm1(fcst[col])

sf.plot(df, fcst, level=[70, 90])
```
Screenshot 2026-05-29 at 8 42 47 p m

eskayML:
Alright that makes sense

Thanks for the careful explanation

PR #1128, opened 2026-04-09
feat: add py.typed marker for PEP 561 compliance

kalichiruvelli-art:
## Summary

Adds an empty `py.typed` marker file to the `statsforecast/` package directory as specified by [PEP 561](https://peps.python.org/pep-0561/).

## Problem

Type checkers (Pyright, Pylance, mypy) raise the following warning when importing `statsforecast`:

```
Stub file not found for "statsforecast"  (reportMissingTypeStubs)
```

Without a `py.typed` marker the package is treated as untyped by all PEP 561-compliant tools.

## Fix

Add an empty `statsforecast/py.typed` file. This zero-byte marker tells type checkers that the package ships with inline type annotations and should be type-checked directly. No logic changes, no new dependencies.

Fixes #1121


kalichiruvelli-art:
Hi @jmoralez and team! Just wanted to gently follow up on this PR. It adds an empty `py.typed` marker file (PEP 561) so that type checkers like Pyright and mypy no longer raise warnings when importing `statsforecast`. It's a zero-byte file with no logic changes. Happy to address any feedback — thanks for your consideration!

nasaul:
Thanks for your contribution, however this is at the wrong path, the correct path should be `python/statsforecast` I'll close this one and merge #1149

PR #1145, opened 2026-05-09
[FIX] theta: avoid NaN PIs when residuals burn-in leaves <2 samples

### LINKED ISSUE #1135: theta: NaN forecast intervals when len(y) == 4

jbbqqf:
## Summary

`forecast_theta` produced `NaN` prediction intervals (and a `RuntimeWarning: Degrees of freedom <= 0 for slice`) on very short series. The burn-in slice `residuals[3:]` left 1 sample for `n == 4`, and `np.std(..., ddof=1)` divides by `n-1 == 0`. Fall back to the full residual array on tiny series so PIs stay finite.

Fixes Nixtla/statsforecast#1135 — *theta: NaN forecast intervals when len(y) == 4*

## Context

`thetacalc` discards the first three observations during a warm-up phase, so `forecast_theta` mirrors that with `residuals[3:]`. The burn-in is an optimisation, not a correctness requirement: with `n == 4` only one residual remains, which makes the unbiased estimator `np.std(..., ddof=1)` undefined. The result is that every `lo-*` / `hi-*` column becomes NaN, breaking downstream code that assumes finite intervals.

## Changes

- `python/statsforecast/theta.py` — guard the burn-in slice. When the residual array has 4 or fewer samples, use the full array (still with `ddof=1`) so a 2-sample std is well-defined. The behaviour is unchanged on every series long enough to leave the burn-in slice usable. The branch carries an inline comment so a reviewer reading the diff cold can see why this fallback is safe.
- `tests/test_theta.py` — add `test_short_series_finite_intervals[n=4|5|6]` parametrised regression that asserts the PIs are finite and have positive width for these short-series cases.

```bash
git clone https://github.com/Nixtla/statsforecast.git /tmp/sf-1135 && cd /tmp/sf-1135

REPRO='
import warnings, numpy as np
from statsforecast.models import AutoTheta
warnings.simplefilter("error", RuntimeWarning)  # surface the buggy warning
for n in (4, 5, 6, 8, 50):
    y = (np.arange(n).astype(float) * 0.3
         + np.random.default_rng(0).standard_normal(n) * 0.2 + 10.0)
    out = AutoTheta(season_length=1).forecast(y, h=3, level=[95])
    w = float(out["hi-95"][0] - out["lo-95"][0])
    assert np.isfinite(w), f"NaN width at n={n}"
    print(f"n={n:>3}  width@h=1 = {w:.6f}")
'

git checkout origin/main
python -c "$REPRO"
# Expected: AssertionError "NaN width at n=4" (RuntimeWarning was raised first)

# --- AFTER (this PR) ---
git fetch https://github.com/jbbqqf/statsforecast.git feat/1135-theta-nan-pi
git checkout FETCH_HEAD
python -c "$REPRO"
# Expected: prints widths for n=4..50 with no NaN, no RuntimeWarning
```

## What I ran locally

- `pytest tests/test_theta.py -q --no-cov` → 23/23 passed (was 22 on origin/main; 23rd is the new regression).
- Regression check: stashing the source change makes `tests/test_theta.py::TestThetaModel::test_short_series_finite_intervals[4]` fail on origin/main with the exact `RuntimeWarning: Degrees of freedom <= 0 for slice` from the bug report; n=5 and n=6 already pass on main and continue to pass after the fix.

## Edge cases tested

| 1 | `n == 4` (the report) | random series of length 4, `level=[95]` | finite, positive-width PIs | `test_short_series_finite_intervals[4]` |
| 2 | `n == 5` (boundary, where burn-in leaves exactly 2 samples) | length-5 series | finite, positive-width PIs (unchanged behaviour) | `test_short_series_finite_intervals[5]` |
| 3 | `n == 6` (well above burn-in) | length-6 series | finite, positive-width PIs (unchanged behaviour) | `test_short_series_finite_intervals[6]` |
| 4 | Long series (regression for behaviour preservation) | AirPassengers (`n == 144`) | identical PIs to origin/main | existing `test_optimal_theta_forecasts`, `test_simple_theta_forecasts`, `test_theta_initialization_functions` |

## Risk / blast radius

Additive — the new branch only activates when `residuals.size <= 4`. On any series with `n >= 5` the original `residuals[3:]` slice is taken verbatim and `sigma` is bit-identical to `main`. No public API changes.

## Release note

```release-note
fix(theta): avoid NaN prediction intervals when the input series is shorter than the residual burn-in (n <= 4)
```

---

*PR drafted with assistance from Claude Code. The change was reviewed manually against `python/statsforecast/theta.py` and the matching burn-in choice in `thetacalc`. The reproducer block above was used during development; it is the same one a reviewer can paste verbatim.*

### DIFF
--- a/python/statsforecast/theta.py
+++ b/python/statsforecast/theta.py
@@ -333,7 +333,16 @@ def forecast_theta(obj, h, level=None):
     res = {"mean": forecast}
 
     if level is not None:
-        sigma = np.std(obj["residuals"][3:], ddof=1)
+        # The first 3 residuals are a burn-in matching `thetacalc`'s warm-up
+        # phase. For very short series (n <= 4) the burn-in leaves <2 samples,
+        # so std with ddof=1 yields NaN. Fall back to all residuals to keep
+        # PIs finite — the burn-in is an optimisation, not a correctness
+        # requirement, and on tiny series every observation matters.
+        residuals = obj["residuals"]
+        if residuals.size > 4:
+            sigma = np.std(residuals[3:], ddof=1)
+        else:
+            sigma = np.std(residuals, ddof=1)
         mean_y = obj["mean_y"]
         samples = compute_pi_samples(
             n=n,
--- a/tests/test_theta.py
+++ b/tests/test_theta.py
@@ -504,3 +504,23 @@ class TestThetaModel:
         res_transfer = forward_theta(res, intermittent_series)
         for key in res_transfer["par"]:
             assert res["par"][key] == res_transfer["par"][key]
+
+    @pytest.mark.parametrize("n", [4, 5, 6])
+    def test_short_series_finite_intervals(self, n):
+        """Regression for #1135: AutoTheta PIs were NaN for n == 4 because the
+        residual burn-in `[3:]` left 1 sample, and ``np.std`` with ``ddof=1``
+        divided by zero. The fix falls back to all residuals when the burn-in
+        would leave too few. PIs must stay finite (and non-degenerate) on
+        short series.
+        """
+        from statsforecast.models import AutoTheta
+
+        rng = np.random.default_rng(0)
+        y = (np.arange(n).astype(float) * 0.3
+             + rng.standard_normal(n) * 0.2 + 10.0)
+        out = AutoTheta(season_length=1).forecast(y, h=3, level=[95])
+        assert np.all(np.isfinite(out["lo-95"])), out
+        assert np.all(np.isfinite(out["hi-95"])), out
+        # Width should be strictly positive — a degenerate zero-width interval
+        # would mean sigma collapsed to ~0.
+        assert (out["hi-95"] - out["lo-95"] > 0).all(), out


nasaul:
Already solved in previous PR, thanks for your contribution

PR #1158, opened 2026-06-01
[FEAT]: Unobserved Components Model native implementation in StatsForecast

MMenchero:
## Description

Currently, the Unobserved Components Model (UCM) in `statsforecast` is a wrapper of the implementation in `statsmodels`. This PR introduces a native implementation that produces very similar results to `statsmodels` with the efficiency of `statsforecast`. 

This is a WIP and for now it includes a minimal UCM model with a local trend + seasonal + residual. Other versions will be added soon.


python/statsforecast/models.py
```diff
@@ -6285,6 +6285,142 @@ def __init__(
...
+            res["fitted"] = fcst["fitted"]
+        if level is not None:
+            level = sorted(level)
+            if self.prediction_intervals is not None:
+                res = self._add_conformal_intervals(fcst=res, y=y, X=X, level=level)
+            else:
+                res = {**res, **_calculate_intervals(res, level, h, fcst["sigma"])}
+            if fitted:
+                se = _calculate_sigma(y - fcst["fitted"], len(y))
+                res = _add_fitted_pi(res=res, se=se, level=level)
```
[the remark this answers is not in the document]
MMenchero: This is a false positive as `se` is a NumPy scalar, not a Python float. Plus all the checks have passed.

python/statsforecast/models.py
```diff
@@ -6285,6 +6285,142 @@ def __init__(
...
+            level (List[float]): Confidence levels (0-100) for prediction intervals.
+
+        Returns:
+            dict: Dictionary with entries `fitted` for point predictions and `level_*` for probabilistic predictions.
+        """
+        res = {"fitted": self.model_["fitted"]}
+        if level is not None:
+            level = sorted(level)
+            res = _add_fitted_pi(res=res, se=self.model_["sigma"], level=level)
+        return res
```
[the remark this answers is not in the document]
MMenchero: This is a false positive as `se` is a NumPy scalar, not a Python float. Plus all the checks have passed.

tests/test_ucm.py
```diff
@@ -1,249 +1,304 @@
...
-    from statsforecast.models import UCM
-
-    np.random.seed(42)
-    y = np.cumsum(np.random.randn(100)) + 50
-
-    model = UCM(level='local level')
-    model.fit(y)
-
-    forecast = model.predict(h=10)
+warnings.simplefilter("ignore")
```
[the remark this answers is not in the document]
MMenchero: This is the standard `statsforecast` convention. Two other test files, `test_arima.py` and `test_garch.py` also use `import warnings` and `warnings.simplefilter("ignore")` at the top.

python/statsforecast/ucm.py
```diff
@@ -1,549 +1,197 @@
...
+    bounds = [(1e-8 * var_y, None)] * n_params
+
+    result = minimize(
+        _nll,
+        x0,
+        args=(y, Z, T, R, a0, P0, has_seasonal),
+        method="L-BFGS-B",
+        bounds=bounds,
+    )
+    params = result.x
```
nasaul, line 163: We should add a warning in the optimizer:

``` python
if not result.success:
      import warnings
      warnings.warn(f"UCM optimization did not converge:
  {result.message}", UserWarning)
```

python/statsforecast/ucm.py
```diff
@@ -1,549 +1,197 @@
...
+    RQRt = R @ Q @ R.T
+
+    a_pred = a0.astype(np.float64).copy()
+    P_pred = P0.astype(np.float64).copy()
+
+    one_step_pred = np.empty(n, dtype=np.float64)
+    loglik = 0.0
+    log_2pi = np.log(2.0 * np.pi)
+
+    for t in range(n):
```
nasaul, line 77: This Python loop could become super slow. Not something that we should tackle now, but maybe a good idea to eventually convert this loop to C++.

python/statsforecast/models.py
```diff
@@ -6285,6 +6285,143 @@ def __init__(
...
+        - [Harvey, A. C. (1989). "Forecasting, Structural Time Series Models and the Kalman Filter". Cambridge University Press.](https://www.cambridge.org/core/books/forecasting-structural-time-series-models-and-the-kalman-filter/CE5E112570A56960601760E786A5E631)
+        - [Durbin, J. and Koopman, S. J. (2012). "Time Series Analysis by State Space Methods". 2nd ed. Oxford University Press.](https://academic.oup.com/book/16563)
+
+    Args:
+        season_length (int): Number of observations per seasonal cycle. Ex: 12 for monthly data. Use 1 for a trend-only model.
+        alias (str): Custom name of the model.
+        prediction_intervals (Optional[ConformalIntervals]): Information to compute conformal prediction intervals. By default, the model computes parametric (Gaussian) prediction intervals.
+    """
+
+    def __init__(
```
nasaul, line 6313: I think that we should use the same signature as [before](https://github.com/Nixtla/statsforecast/blob/main/python/statsforecast/ucm.py#L48) to eventually converge into the same behaviour.

python/statsforecast/ucm.py
```diff
@@ -1,549 +1,197 @@
...
+        v = y[t] - y_hat
+        Pz = P_pred @ z
+        F = z @ Pz + H
+
+        # Update step (skip if F is not usable, e.g. during diffuse start).
+        if np.isfinite(F) and F > 0:
+            loglik += -0.5 * (log_2pi + np.log(F) + v * v / F)
+            K = Pz / F
+            a_filt = a_pred + K * v
+            P_filt = P_pred - np.outer(K, Pz)
```
nasaul, line 90: In statsmodels they use: [`P_filt = 0.5*(P_pred - np.outer(K, Pz))`](https://github.com/statsmodels/statsmodels/blob/main/statsmodels/tsa/statespace/_kalman_filter.pyx.in#L1258) to ensure symmetry, I suggest doing the same.

nasaul (commented):
Great work Mariana! I think that we need a few changes before merging.

python/statsforecast/ucm.py
MMenchero: Added.

python/statsforecast/ucm.py
MMenchero: Agreed, I'll keep this in mind, but first I'll do a comparison of the current version vs `statsmodels` using the M3 and M4 datasets to have an idea of how it performs in Python.

python/statsforecast/models.py
MMenchero: I understand why you made this suggestion, and the goal is to eventually support all the features available in `statsmodels`. However, since this is a native implementation, I’d prefer not use the same signature yet, as some features may be better implemented differently.
    
The downside is that we may need to wait until more features are implemented before merging UCM, since this will be a breaking change for users. Still, I think it's better if we retain the current signature and don't constraint our implementation choices. Eventually, it will probably converge to something very similar to `statsmodels`.

python/statsforecast/ucm.py
MMenchero: I ended up doing this since the experiments with the M3 dataset using the Nelder-Mead optimization method show that `statsforecast` was as accurate as `statsmodels`, but slightly slower. The C++ implementation made `statsforecast` as fast or even faster than `statsmodels`. 

**Results for M3**
| dataset      | model   |     mae |    rmse |   smape |   time |
|:-------------|:--------|--------:|--------:|--------:|-------:|
| M3_Monthly   | UCM-sf  |  700.61 |  836.78 |    8.11 |   0.18 |
| M3_Monthly   | UCM-sm  |  694.52 |  832.49 |    8.23 |   0.27 |
| M3_Other     | UCM-sf  |  261.94 |  292.9  |    2.42 |   0.04 |
| M3_Other     | UCM-sm  |  245.82 |  282.72 |    2.73 |   0.05 |
| M3_Quarterly | UCM-sf  |  573.46 |  671.75 |    5.56 |   0.06 |
| M3_Quarterly | UCM-sm  |  599.4  |  699.82 |    5.9  |   0.08 |
| M3_Yearly    | UCM-sf  | 1331.58 | 1524.22 |   10.14 |   0.06 |
| M3_Yearly    | UCM-sm  | 1378.62 | 1576.04 |   10.61 |   0.06 |

PR #1165, opened 2026-06-15
add conformal_error prediction intervals method

### LINKED ISSUE #1164: [FEAT] Add conformal_error method for prediction intervals
Hey, I noticed `conformal_error` is commented out in `_get_conformal_method` and never implemented. The distribution-based method is there but the simpler error-based one is missing.

The idea is straightforward: take the quantile of absolute residuals at the desired confidence level, then do mean ± quantile(cs, level/100) for symmetric intervals. It's maybe 15 lines of actual logic across models.py and utils.py.

I can put together a PR with implementation + tests if you're interested. Just wanted to check first since there might be a reason it was left commented out.

DresdenGman:
Implements the conformal_error method for prediction intervals. This uses quantiles of conformity scores directly as error margins, as a simpler alternative to conformal_distribution.

Closes #1164


nasaul:
Hello @DresdenGman I think that this is addition is a good thing, however there are a lot changes that doesn't need to be made in order to make this functionality available. Please clean up this PR.

DresdenGman:
Closing in favor of #1175 — a clean rebuild with only the conformal_error changes (no unrelated formatting or notebook markers).

PR #1175, opened 2026-07-04
[FEAT] Add conformal_error prediction intervals method

DresdenGman:
Adds the `conformal_error` prediction intervals method, which uses quantiles of the conformity scores as the error margin for prediction intervals. This is a companion to the existing `conformal_distribution` method.

Changes:
- Add `_add_conformal_error_intervals()` function to models.py
- Enable `"conformal_error"` in `_get_conformal_method()`
- Add test `test_conformal_error_intervals`

### DIFF
--- a/python/statsforecast/models.py
+++ b/python/statsforecast/models.py
@@ -149,8 +149,9 @@ def _add_conformal_error_intervals(
 ) -> Dict:
     r"""
     Adds conformal intervals to the `fcst` dict based on conformal scores `cs`.
-    `level` should be already sorted. This strategy uses the quantiles of the
-    conformity scores as the error margin for the prediction intervals.
+    `level` should be already sorted. Uses the `lv / 100` quantile of the
+    absolute conformity scores as the error margin around `fcst["mean"]`.
+    Returns the modified `fcst` dict.
     """
     quantiles = {lv: np.quantile(cs, lv / 100, axis=0) for lv in level}
     for lv in reversed(level):
@@ -165,10 +166,10 @@ def _get_conformal_method(method: str):
         "conformal_distribution": _add_conformal_distribution_intervals,
         "conformal_error": _add_conformal_error_intervals,
     }
-    if method not in available_methods.keys():
+    if method not in available_methods:
         raise ValueError(
             f"prediction intervals method {method} not supported "
-            f"please choose one of {', '.join(available_methods.keys())}"
+            f"please choose one of {', '.join(available_methods)}"
         )
     return available_methods[method]
 
--- a/python/statsforecast/models.py
+++ b/python/statsforecast/models.py
@@ -142,10 +142,28 @@ def _add_conformal_distribution_intervals(
     return fcst
 
 
+def _add_conformal_error_intervals(
+    fcst: Dict,
+    cs: np.ndarray,
+    level: List[Union[int, float]],
+) -> Dict:
+    r"""
+    Adds conformal intervals to the `fcst` dict based on conformal scores `cs`.
+    `level` should be already sorted. This strategy uses the quantiles of the
+    conformity scores as the error margin for the prediction intervals.
+    """
+    for lv in level:
+        alpha = 100 - lv
+        margin = np.quantile(cs, 1 - alpha / 200, axis=0)
+        fcst[f"lo-{lv}"] = fcst["mean"] - margin
+        fcst[f"hi-{lv}"] = fcst["mean"] + margin
+    return fcst
+
+
 def _get_conformal_method(method: str):
     available_methods = {
         "conformal_distribution": _add_conformal_distribution_intervals,
-        # "conformal_error": _add_conformal_error_intervals,
+        "conformal_error": _add_conformal_error_intervals,
     }
     if method not in available_methods.keys():
         raise ValueError(
--- a/python/statsforecast/models.py
+++ b/python/statsforecast/models.py
@@ -152,11 +152,11 @@ def _add_conformal_error_intervals(
     `level` should be already sorted. This strategy uses the quantiles of the
     conformity scores as the error margin for the prediction intervals.
     """
+    quantiles = {lv: np.quantile(cs, lv / 100, axis=0) for lv in level}
+    for lv in reversed(level):
+        fcst[f"lo-{lv}"] = fcst["mean"] - quantiles[lv]
     for lv in level:
-        alpha = 100 - lv
-        margin = np.quantile(cs, 1 - alpha / 200, axis=0)
-        fcst[f"lo-{lv}"] = fcst["mean"] - margin
-        fcst[f"hi-{lv}"] = fcst["mean"] + margin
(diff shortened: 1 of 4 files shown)


python/statsforecast/models.py
```diff
@@ -142,10 +142,28 @@ def _add_conformal_distribution_intervals(
...
+    level: List[Union[int, float]],
+) -> Dict:
+    r"""
+    Adds conformal intervals to the `fcst` dict based on conformal scores `cs`.
+    `level` should be already sorted. This strategy uses the quantiles of the
+    conformity scores as the error margin for the prediction intervals.
+    """
+    for lv in level:
+        alpha = 100 - lv
+        margin = np.quantile(cs, 1 - alpha / 200, axis=0)
```
nasaul: The quantile cut here is over-conservative and inconsistent with the rest of the Nixtla ecosystem. `cs` is already the *absolute* error (`|forecast - actual|`, see `_conformity_scores`), so standard split conformal takes the `level/100` quantile — the 90th percentile for a 90% interval. `1 - alpha/200` takes the 95th percentile instead.

Two references agree on `lv/100`:
- mlforecast's `_add_conformal_error_intervals` uses `cuts = [lv / 100 for lv in level]`.
- This repo's `conformal_distribution` builds symmetric paths `mean ± cs` and takes the `1 - alpha/200` quantile of that doubled sample, which is mathematically equivalent to the `lv/100` quantile of `|cs|`.

As written, `conformal_error` would systematically produce wider intervals than `conformal_distribution` on the same scores.

```suggestion
        margin = np.quantile(cs, lv / 100, axis=0)
```

python/statsforecast/models.py
```diff
@@ -142,10 +142,28 @@ def _add_conformal_distribution_intervals(
...
+    fcst: Dict,
+    cs: np.ndarray,
+    level: List[Union[int, float]],
+) -> Dict:
+    r"""
+    Adds conformal intervals to the `fcst` dict based on conformal scores `cs`.
+    `level` should be already sorted. This strategy uses the quantiles of the
+    conformity scores as the error margin for the prediction intervals.
+    """
+    for lv in level:
```
nasaul, line 159: Key insertion order doesn't match the repo convention, and dict order propagates to the output columns. `_add_conformal_distribution_intervals` (and the existing `test_conformal_intervals`) emit all `lo` columns in descending level order, then all `hi` ascending: `["mean", "lo-90", "lo-80", "hi-80", "hi-90"]`. This loop produces `["mean", "lo-80", "hi-80", "lo-90", "hi-90"]` — I verified this empirically, and it's the second reason the new test fails even after fixing the `utils.py` gate.

Something like:

```python
quantiles = {lv: np.quantile(cs, lv / 100, axis=0) for lv in level}
for lv in reversed(level):
    fcst[f"lo-{lv}"] = fcst["mean"] - quantiles[lv]
for lv in level:
    fcst[f"hi-{lv}"] = fcst["mean"] + quantiles[lv]
```

tests/test_models.py
```diff
@@ -143,6 +143,20 @@ def test_conformal_intervals():
     assert list(fcst_conformal.keys()) == ["mean", "lo-90", "lo-80", "hi-80", "hi-90"]
 
 
+def test_conformal_error_intervals():
+    """Test conformal_error method produces valid intervals."""
+    conf_intervals = ConformalIntervals(h=12, n_windows=10, method="conformal_error")
```
nasaul: This test currently fails: `ConformalIntervals.__init__` in `python/statsforecast/utils.py` still restricts `allowed_methods` to `["conformal_distribution"]`, so this constructor raises `ValueError` before the method is ever reached. That file needs to be updated as part of this PR.

tests/test_models.py
```diff
@@ -143,6 +143,20 @@ def test_conformal_intervals():
...
+    zero_model = ZeroModel(conf_intervals)
+    fcst = zero_model.forecast(ap, h=12, level=[80, 90])
+    # Check keys
+    assert list(fcst.keys()) == ["mean", "lo-90", "lo-80", "hi-80", "hi-90"]
+    # For ZeroModel (mean=0), intervals should be symmetric around 0
+    np.testing.assert_array_equal(fcst["lo-90"], -fcst["hi-90"])
+    np.testing.assert_array_equal(fcst["lo-80"], -fcst["hi-80"])
+    # Higher confidence = wider intervals
+    assert np.all(fcst["hi-90"] >= fcst["hi-80"])
```
nasaul, line 157: The symmetry and monotonicity assertions hold for almost any symmetric-margin construction, so they wouldn't catch the quantile-convention issue flagged in `models.py`. Consider adding an exact-value check, e.g. `np.testing.assert_allclose(fcst["hi-90"], np.quantile(cs, 0.9, axis=0))` on a known score matrix — or asserting the ZeroModel intervals match `conformal_distribution`'s output, since the two methods should coincide in this symmetric case.

nasaul (changes requested):
Thanks for picking this up. However, there's a blocker outside the diff plus two issues in the new code (inline comments below).

**Blocker: the method can't actually be enabled.** `ConformalIntervals.__init__` in `python/statsforecast/utils.py` still has `allowed_methods = ["conformal_distribution"]`, so `ConformalIntervals(method="conformal_error")` — the only public entry point, and what the new test constructs — raises `ValueError`. I ran `test_conformal_error_intervals` locally and it fails at exactly that line. Please add `"conformal_error"` to `allowed_methods` and update the class docstring there as well.

Given that gate was missed, it would also be worth adding one end-to-end test through a real model (e.g. `AutoARIMA(prediction_intervals=ConformalIntervals(method="conformal_error"))` via `StatsForecast.forecast`) so the public path is exercised.

DresdenGman:
Thanks for the thorough review — all four points addressed:

- **allowed_methods gate**: added `"conformal_error"` to `ConformalIntervals.__init__` and updated the class docstring.
- **Quantile cut**: changed from `1 - alpha / 200` to `lv / 100`. On absolute scores this is the correct split-conformal cut, and it brings the method in line with mlforecast and the equivalent path through `conformal_distribution`.
- **Column ordering**: switched to compute all quantiles first, then emit `lo` in descending level order and `hi` in ascending — matches `conformal_distribution`'s output convention.
- **Test strength**: added exact-value `assert_allclose` against `np.quantile(cs, 0.9/0.8, axis=0)`. These would have caught the quantile formula bug even without the visual key-order assertion.

I haven't added an end-to-end test through `AutoARIMA` + `StatsForecast.forecast` yet — let me know if you'd like that in this PR or as a follow-up.

nasaul:
Hello @DresdenGman have you pushed the changes? I cannot see anything from this side.

DresdenGman:
Sorry about that — the fix commits were on a different branch. Just force-pushed them to the PR branch — you should see 3 new commits from 2026-07-10 now.

nasaul (approved):
Thanks for your contribution. Happy forecasting

PR #1176, opened 2026-07-04
Add py.typed marker

DresdenGman:
Closes #1121.

This adds a `py.typed` marker to the `statsforecast` package so type checkers can recognize it as a typed package following PEP 561.

I verified that the marker is included in both the sdist and wheel:

- sdist: `statsforecast-0.0.0/python/statsforecast/py.typed`
- wheel: `statsforecast/py.typed`

No runtime code is changed.

### DIFF
--- a/.github/workflows/build-docs.yaml
+++ b/.github/workflows/build-docs.yaml
@@ -35,7 +35,7 @@ jobs:
           python-version: "3.10"
 
       - name: Install uv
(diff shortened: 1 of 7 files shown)


MMenchero:
This is ready for merging @nasaul.

nasaul (approved):
Thanks :)

PR #1195, opened 2026-07-22
migrate to scikit-build-core

jmoralez:
Migrates from setup.py with setuptools to pyproject.toml with scikit-build-core


jmoralez:
@codspeedbot explain why this is faster

nasaul (dismissed):
Approving — packaging completeness verified against real builds on macOS/arm64. One nit I'd like addressed before merge.

## Required nit: exclude stale build artifacts from the wheel

`wheel.exclude` currently only holds `["Release"]`, but `wheel.packages = ["python/statsforecast"]` copies that directory wholesale. Any stale in-place extension left in the source tree gets shipped:

```toml
wheel.exclude = ["Release", "**/*.so", "**/*.pyd", "**/*.dylib"]
```

Reproduced with three leftover `_lib.cpython-{39,312,313}-darwin.so` files in the tree — all three landed in a `cp313` wheel. To be clear, **this is not a regression**: `main`'s setuptools build (`package-data "*" = ["**/*"]`) does exactly the same thing, and CI is unaffected since cibuildwheel builds from clean checkouts. But now that `setup.py build_ext --inplace` is gone from the dev workflow, contributors with older trees are the likely victims, so it's a cheap guard to add here.

## Verified

Everything from the old `MANIFEST.in` + `package-data` still ships. Wheel file list is an exact match for the source tree:

- `py.typed` present — this was the one I most expected to be silently dropped
- `adapters/` and `distributed/` subpackages present (neither was ever explicitly declared, so behavior is unchanged)
- `_lib` built from all 9 `src/*.cpp`, exposing `arima, ces, ets, garch, mfles, ses, tbats, theta`
- `version` is static, so no `metadata.version.provider` needed
- `settings.ini` in the old `MANIFEST.in` doesn't exist in the repo — stale entry, fine to drop

sdist is the risky part given `sdist.inclusion-mode = "explicit"`, so I built a wheel **from the extracted sdist**: identical file list, Eigen submodule bundled (2.7 MB tarball, 1785 files). All workflows use `submodules: "true"`, so sdist and cibuildwheel jobs will have Eigen.

API surface, checked against the **installed wheel** rather than the source tree:

| Check | Result |
|---|---|
| Submodule imports | 20/21 (only `adapters.prophet`, an intentional optional-dep guard) |
| `models.__all__` | 41/41 classes, each with `fit`/`predict`/`forecast` |
| `StatsForecast` | all 11 public methods |
| `MultiprocessBackend` / `FugueBackend` | both subclass `ParallelBackend`, both expose `forecast` + `cross_validation` |

- Clean-worktree wheel build: 7.5s, no warnings
- `pytest tests/` against the installed wheel: **604 passed**
- `tests/test_distributed_fugue.py` with dask + ray + pyspark: **6 passed**
- End-to-end `forecast(h=6, level=[80])` and `cross_validation` through the compiled extension: correct shapes, sane values
- Grepped `.github/`, `Makefile`, `action_files/` for `setup.py` / `MANIFEST.in` / `build_ext` — no stale references; the `CONTRIBUTING.md` line is fixed here

## Unrelated, take it or leave it

`python/statsforecast/adapters/prophet.py:15` and `:26` are missing an `f` prefix, so users see a literal `{e}` in the "install prophet" error. Pre-existing and out of scope for this PR — just noting it since the import surfaced it.

jmoralez:
Thanks for the thorough pass — the nit is real and reproduces on linux/aarch64 too. I fixed it in c2eeae9, but via `sdist.exclude` rather than `wheel.exclude`, because the suggested patch produces a broken wheel.

### Why `wheel.exclude` doesn't work here

`wheel.exclude` isn't a filter on the package copy — it's applied at zip time across the whole platlib tree ([`_wheelfile.py:223-239`](https://github.com/scikit-build/scikit-build-core/blob/main/src/scikit_build_core/build/_wheelfile.py)), which includes the CMake install output. With `wheel.install-dir = "statsforecast"`, the stale artifacts and the real `_lib` land in the *same* directory, so no pattern can separate them.

Building with exactly the suggested config:

```
binary members: (NONE)          # was 5
$ python -c "from statsforecast.models import AutoARIMA"
ModuleNotFoundError: No module named 'statsforecast._lib'
```

### What I did instead

```toml
sdist.exclude = ["**/__pycache__", "**/*.pyc", "**/*.egg-info", "**/*.so", "**/*.pyd", "**/*.dylib"]
```

`sdist.exclude` is passed as `src_exclude` into `packages_to_file_mapping`, so it filters the source tree during the package copy and never sees CMake output.

Worth adding to your root-cause note: the reason this repo is exposed at all is `sdist.inclusion-mode = "explicit"`, which makes scikit-build-core skip `.gitignore` entirely (`"manual" and "explicit" do not consult git ignore files at all`). We do have `*.so` in `.gitignore`, it just buys us nothing here.

### Verified

Planted stale `_lib.cpython-{39,313}-darwin.so`, `_lib.pyd`, `_lib.dylib` and diffed against a baseline build:

| | dropped | added |
|---|---|---|
| wheel | the 4 planted artifacts | none |
| sdist | those 4 **+ `_lib.cpython-312-aarch64-linux-gnu.so`** | none |

That sdist row is the bit `wheel.exclude` would have missed regardless: the sdist was leaking binaries too, including the current-platform one, since `sdist.exclude` sits on that path as well.

Fixed wheel installs clean, `_lib` exposes all 8 submodules, and `AutoARIMA`/`AutoETS`/`AutoTheta` `forecast(h=6, level=[80])` all return sane values through the extension.

Leaving the `adapters/prophet.py` f-string nit out of this PR as you suggested — it's real (lines 14 and 25), just unrelated to the build migration.

nasaul (approved):
Re-approving on `c2eeae9` — my earlier approval was dismissed when this commit landed.

## The nit is resolved

I asked for `**/*.so` / `**/*.pyd` / `**/*.dylib` in `wheel.exclude`; they went into `sdist.exclude` instead. That's sufficient, and it's the better spelling: `build/wheel.py` passes `src_exclude=settings.sdist.exclude` into `packages_to_file_mapping`, so the source-side walk of `wheel.packages` is filtered by `sdist.exclude` — `wheel.exclude` only filters the target side. One list covers both artifacts.

Re-verified on this commit with a planted stale in-place `python/statsforecast/_lib*.so`: it stays out of the sdist and out of the wheel, and the editable meta-path finder shadows it rather than resolving to it. Also rebuilt from the extracted sdist (39 files, exact) and confirmed the wheel layout — `statsforecast/*.py`, `py.typed`, CMake-installed `_lib*.so`.

## Non-blocking follow-ups

None of these block the migration; flagging for a later pass.

1. **`pyproject.toml:88` — `setuptools<84` dropped from the `all` group.** Probably incidental: it's still in `dev`. On the 3.14 legs of `pytest.yaml` (`uv sync --group all`) nothing pulls setuptools in anymore, since its only remaining requirer is `dash` via `plotly-resampler ; python_version < '3.14'`. CI is green today, so this is latent rather than broken.
2. **`pyproject.toml:6` — `build-dir = "build/{wheel_tag}"` gives no incremental compilation under build isolation.** scikit-build-core drops `CMakeCache.txt` and `CMakeFiles/` (ninja's `.o` files) whenever `skbuild_path` changes, which it does on every isolated build. So the rebuild command we now document in CONTRIBUTING (`uv sync --group dev --reinstall-package statsforecast`) is a full 9-TU + LTO rebuild every time, where setuptools' `naive_recompile` plus a persistent `build/temp.*` was not. Worth a documented non-isolated path for contributors.
3. **`CMakeLists.txt:5` — `CMAKE_CXX_STANDARD 20` without `CMAKE_CXX_STANDARD_REQUIRED ON` / `CMAKE_CXX_EXTENSIONS OFF`.** A pre-C++20 compiler silently downgrades instead of failing, and the sources need `<span>` and `<numbers>`. Also means we build `-std=gnu++20` where setup.py used a strict `-std=c++20`.
4. **`pyproject.toml:7` — `build.verbose = true` / `logging.level = "INFO"` ship in the published config**, so every end-user source build prints full compiler command lines. These read like CI knobs; `SKBUILD_BUILD_VERBOSE` and `SKBUILD_LOGGING_LEVEL` in the workflow would keep them out of users' builds.

`wheel.exclude = ["Release"]` is now vestigial (matches nothing in either tree) but harmless.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1206, opened 2026-08-11
arima: fix coef stderr

jmoralez:


nasaul:
## Review: the negative variances, the m3 change, and the shared root cause

All of it traces back to one thing: **`mod` is mutated in place by the objective and never fully restored**, so the ML objective is not a function of its parameters.

### 1. What gets mutated through `minimize`

Args bound into the new partials: `x, coef, mask, arma, mod, ncxreg, xreg, narma, trans`.

| Variable | Mutated by the objective? |
|---|---|
| **`mod`** | **Yes — in place, and not fully restored** |
| `x` | No (`x = x.copy()` first), but mutated in the *enclosing* scope: `x -= xreg @ coef[...]` (`arima.py:686`, `arima.py:885`) |
| `coef` | No (`par = coef.copy()`), but mutated in enclosing scope (`coef[mask] = res.x`, `maInvert`, then rebound by `arima_undopars`) |
| `init` | Mutated in place at `arima.py:713` — and it is the caller's array when the user passes `init=` |
| `mask, arma, xreg, ncxreg, narma, trans` | Read-only |

`upARIMA` (`arima.py:226`) writes `mod["phi"]`, `mod["theta"]`, `mod["T"][:p,0]`, `mod["Pn"][:r,:r]`, `mod["a"][:] = 0` and returns **the same dict**. `arima_like` then passes `mod["a"]`, `mod["P"].ravel()`, `mod["Pn"].ravel()` — all views — into C++, which writes through them (`src/arima.cpp:405`, `src/arima.cpp:412`).

`upARIMA` restores only `a` and the **top-left `r×r`** of `Pn`. The diffuse block `Pn[r:, r:]` (set to `kappa = 1e6` in `make_arima`) is never restored, and the filter overwrites it. For `(2,1,1)(0,1,0)[12]`:

```
Pn diffuse diag BEFORE any eval : [1e6, 1e6, 1e6, 1e6]
eval 1 (fresh mod)              : 2.4804968998826187
Pn diffuse diag AFTER           : [-1.3e-19, 6.8e-17, 1.8e-16, -2.0e-16]
eval 2 (same mod)               : 2.8883488337137626
eval 3 (same mod)               : 2.8883488337137560
eval 4 (fresh mod again)        : 2.4804968998826187
```

The first evaluation differs from every subsequent one by 0.41. This is pre-existing on `main`; this PR makes it observable.

### 2. Why the variance goes negative

```python
from statsforecast.models import ARIMA
m = ARIMA(order=(2, 1, 1), seasonal_order=(0, 1, 0), season_length=12).fit(ap).model_
```

1. BFGS returns `nit=0`, `status=2` ("precision loss") — it never leaves the CSS warm start, because `f(x₀) = 2.4805` (fresh `mod`) and `f(x₀+h) = 2.8883` (polluted `mod`), so the apparent gradient is ~`0.41/h` and the line search fails immediately.
2. The PR computes a genuine numerical Hessian at that non-stationary point. It is **indefinite**, and stably so across step sizes — not an `epsilon` artifact:

```
eig(H, eps=1e-3) : [-1.86028, -0.02483, +0.73600]
eig(H, eps=1e-2) : [-1.86251, -0.02482, +0.73562]
eig(H, eps=1e-4) : [-1.86041, -0.02480, +0.73603]
```

3. `cho_factor(res.hess)` raises `LinAlgError` (negative leading minor), the `except` falls through to `lu_factor`/`lu_solve`, which solves happily with an indefinite matrix → `var diag = [0.00767, 0.00777, -0.00397]` → `sqrt` → `nan`.

On `main` this was hidden rather than fixed: BFGS's `hess_inv` is PSD by construction, and with `nit=0` it is *exactly the identity*, so `main` returns `var = AᵀA/n_used = [0.00621, 0.00712, 0.00763]` — non-negative and equally meaningless.

Coefficients and AIC are identical on both branches (`aic = 1029.652`); only the variance differs.

### 3. Why the m3 benchmark moved (4.46 → 4.33)

`checkarima` (`arima.py:980`):

```python
return any(np.isnan(np.sqrt(np.diag(obj["var_coef"]))))
```

and `myarima` (`arima.py:1157`): `if minroot < 1.01 or checkarima(fit): fit["ic"] = math.inf`.

Negative variance → NaN SE → **candidate silently dropped from the model search**. Measured on AirPassengers:

```
PR   : selected ARIMA(1,1,0)(0,1,0)[12]  aic 1020.394  candidates 10  rejected 1
main : selected ARIMA(1,1,0)(0,1,0)[12]  aic 1020.394  candidates 10  rejected 0
```

The rejected candidate is exactly `(2,1,1)(0,1,0)[12]`. Same final pick on this series, but across the 3003 m3 series this changes which models survive selection. So the benchmark number did not move because the math got better — it moved because models started being discarded. (I did not re-run m3 itself, only established the mechanism.)

### Recommendations

1. **Fix `upARIMA` to fully restore `mod`** — all of `Pn` including the diffuse `kappa` block, plus `P` — or rebuild `mod` per evaluation. This is the root cause. It makes the objective a pure function of the parameters and lets BFGS actually converge; without it the numerical Hessian is measuring a discontinuous function.
2. **Don't silently fall back to LU.** If `cho_factor` fails, the Hessian is not PD, so the point is not a minimum. Returning NaN (as the C++ path already does, `7781777`) or warning is more honest than inverting an indefinite matrix.
3. **The T / skew-normal / GED branches are a regression.** `approx_hess3(...)[:k,:k]` followed by inversion gives `H_θθ⁻¹`, but the marginal covariance is `[H⁻¹]_θθ = (H_θθ − H_θψ H_ψψ⁻¹ H_ψθ)⁻¹`. Dropping the Schur correction **understates** the ARMA standard errors by ignoring uncertainty in σ²/ν/α/β. The previous `hess_inv[:k,:k]` was the block of the *inverse*, which was correct in that respect.
4. Minor:
   - `np.isnan(res.hess).any()` does not catch `inf`.
   - Hardcoded `epsilon=1e-3` is an *absolute* step; questionable for xreg/intercept coefficients of order 100+.
   - The CSS pre-fit at `arima.py:701` still uses `args=` while everything else moved to `partial`.

PR #1212, opened 2026-08-18
arima: model eligiblity fixes

jmoralez:
Fixes the following:

* Errors on checks for the model eligibility (arCheck and maInvert).
* Adds constraints on the optimization for the new distributions to prevent them from finding points that overflow/underflow and also returns the max float instead of Inf from objective functions so the optimization can continue.
* Removes input mutation in the upARIMA function.
* Sets the correct indentation for the seasonal AR check.

### DIFF
--- a/include/statsforecast/distributions.h
+++ b/include/statsforecast/distributions.h
@@ -20,25 +20,13 @@ inline int distribution_n_extra_params(Distribution d) {
   return (d == Distribution::Laplace) ? 0 : 2;
 }
 
-// Numerically safe box for the optimizer tail: [log_scale, shape].
-//
-// This is THE definition of those limits for the whole project. The Python side
-// reads it through _lib.distributions.tail_bounds() rather than repeating the
-// numbers (see python/statsforecast/distributions.py), so the two can never
-// drift apart.
-//
-// The tail entries are unconstrained by construction (they are logs), so an
-// unbounded line search can propose values that make exp()/lgamma()/pow()
-// overflow or underflow. Observed on macOS-arm64 + numpy>=2 in AutoARIMA+ged:
-// log_beta = -1008 -> exp() underflows to 0.0 -> ZeroDivisionError in
-// lgamma(1.0 / beta) on the Python side.
-//
-// Scale bounds are numeric-safety only (they cannot bind for any real series):
-// they keep exp(x) in [2.7e-109, 3.7e108] and exp(x)**2 in [7.4e-218, 1.4e217]
-// (ged stores log_sigma and reports sigma2 = exp(log_sigma)**2).
-// Shape bounds are statistical: outside them the fitted shape is unusable by
-// scipy's frozen gennorm/t when building prediction intervals.
-inline constexpr double kLogScaleMin = -250.0;
+// Safe box for the optimizer tail. The entries are logs, so an unbounded line
+// search can overflow/underflow exp()/lgamma()/pow(): exp(-1008) == 0.0 then
+// 1.0/beta. Python reads these through _lib.distributions.tail_bounds(), so the
+// two sides cannot drift apart.
+// Scale bounds are numeric-safety only and cannot bind for a real series; shape
+// bounds also keep the fit usable by scipy's gennorm/t.
+inline constexpr double kLogScaleMin = -250.0;  // exp -> [2.7e-109, 3.7e108]
 inline constexpr double kLogScaleMax = 250.0;
 inline constexpr double kLogNuM2Min = -15.0;  // log(nu-2): nu in (2, 1098.6]
 inline constexpr double kLogNuM2Max = 7.0;
@@ -47,8 +35,7 @@ inline constexpr double kAlphaMax = 100.0;
 inline constexpr double kLogBetaMin = -3.0;  // log(beta): beta in [0.05, 50]
 inline constexpr double kLogBetaMax = 3.912023005428146;
 
-// (lo, hi) for each of the two tail entries. Normal/Laplace have no tail, so
-// they get an unbounded box.
+// (lo, hi) per tail entry; Normal/Laplace are unbounded.
 struct TailBounds {
   double scale_lo, scale_hi, shape_lo, shape_hi;
 };
--- a/include/statsforecast/distributions.h
+++ b/include/statsforecast/distributions.h
@@ -20,19 +20,52 @@ inline int distribution_n_extra_params(Distribution d) {
   return (d == Distribution::Laplace) ? 0 : 2;
 }
 
-// Numerically safe box for the optimizer tail, mirroring _DIST_TAIL_BOUNDS in
-// python/statsforecast/distributions.py.  Callers pass these as Nelder-Mead
-// bounds, but the cores clamp too so they stay safe for any caller: outside the
-// box exp()/lgamma()/pow() lose all precision or return inf, and the fitted
-// shape is unusable by scipy's frozen gennorm/t for prediction intervals.
-inline constexpr double kLogScaleMin = -250.0;  // exp(x) in [2.7e-109, 3.7e108]
+// Numerically safe box for the optimizer tail: [log_scale, shape].
+//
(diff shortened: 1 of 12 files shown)


include/statsforecast/distributions.h
```diff
@@ -20,6 +20,44 @@ inline int distribution_n_extra_params(Distribution d) {
   return (d == Distribution::Laplace) ? 0 : 2;
 }
 
+// Safe box for the optimizer tail. The entries are logs, so an unbounded line
```
jmoralez: these are to prevent the optimization from failing due to overflows/underflows

python/statsforecast/arima.py
```diff
@@ -224,21 +232,25 @@ def arima(
     def upARIMA(mod, phi, theta):
         p = len(phi)
         q = len(theta)
-        mod["phi"] = phi
-        mod["theta"] = theta
         r = max(p, q + 1)
-        if p > 0:
-            mod["T"][:p, 0] = phi
+
+        Pn = mod["Pn"].copy()
```
jmoralez: this is to prevent this function from modifying its inputs

python/statsforecast/arima.py
```diff
@@ -447,34 +471,45 @@ def armafn_ged(p_ext, x, trans, coef, mask, arma, mod, ncxreg, xreg, narma):
...
+            + tail_penalty
         )
+        if not math.isfinite(obj):  # e.g. (|e|/sigma)**beta overflowed
+            return np.finfo(np.float64).max
         return obj
 
     def arCheck(ar):
-        p = np.argmax(np.append(1, -ar) != 0)
-        if not p:
+        non_zeros = np.where(np.append(1, -ar) != 0)[0]
```
jmoralez: this check was wrong because argmax returns the first True, which was always the 1 we prepend to the ar and thus it always exited early. the maInvert below had the same problem

python/statsforecast/arima.py
```diff
@@ -447,34 +471,45 @@ def armafn_ged(p_ext, x, trans, coef, mask, arma, mod, ncxreg, xreg, narma):
...
-        if not q0:
+        non_zeros = np.where(np.append(1, ma) != 0)[0]
+        if non_zeros.size == 0:
             return ma
+        q0 = non_zeros[-1]
         coefs = np.append(1, ma[:q0])
         roots = np.polynomial.polynomial.polyroots(coefs)
         ind = np.abs(roots) < 1
-        if any(ind):
+        if not any(ind):
```
jmoralez, line 507: this codition was also wrong

python/statsforecast/arima.py
```diff
@@ -623,15 +658,9 @@ def maInvert(ma):
         if nan_mask.any():
             init[nan_mask] = init0[nan_mask]
         if method == ArimaMethod.ML:
-            # check stationarity
-            if arma[0] > 0:
-                if not arCheck(init[: arma[0]]):
-                    raise ValueError("non-stationary AR part")
-                if arma[2] > 0:
```
jmoralez, line 630: this had the wrong indentation, it's been extracted to a function and fixed

python/statsforecast/arima.py
```diff
@@ -1804,7 +1852,7 @@ def run_tests(x, test, alpha):
         return d
     while dodiff and d < max_d:
         d += 1
-        x = diff(x, 1, 1)[1:]
```
jmoralez, line 1807: this unnecessarily removed an element

tests/test_arima.py
```diff
@@ -889,15 +903,15 @@ def test_issue_649(capsys):
     captured = capsys.readouterr()
     expected_output = """ARIMA(2,0,2)(1,0,1)[12] with non-zero mean : inf
 ARIMA(0,0,0)            with non-zero mean : 494.2237
-ARIMA(1,0,0)(1,0,0)[12] with non-zero mean : inf
-ARIMA(0,0,1)(0,0,1)[12] with non-zero mean : inf
+ARIMA(1,0,0)(1,0,0)[12] with non-zero mean : 496.9135
```
jmoralez, line 909: these are now closer to R (there's still a scaling issue that I'll address in a follow-up PR)

python/statsforecast/arima.py
```diff
@@ -221,24 +233,36 @@ def arima(
...
         p = len(phi)
         q = len(theta)
-        mod["phi"] = phi
-        mod["theta"] = theta
         r = max(p, q + 1)
-        if p > 0:
-            mod["T"][:p, 0] = phi
+
+        if not SSG:
+            raise NotImplementedError('SSinit != "Gardner1980"')
```
nasaul: This `NotImplementedError` for `SSinit != "Gardner1980"` now fires unconditionally, even when `r = max(p, q+1) <= 1`. On `main`, that check only lived inside the `r > 1` branch, so `arima(y, order=(1,0,0), method='ML', SSinit='Rossignol2011')` succeeded (r=1 took the else-branch and never hit the raise). On this branch it now raises for every `r`. Reproduced directly by comparing main vs. this branch.

Suggest gating the raise on `r > 1` (or otherwise excluding the case that previously worked) so non-Gardner1980 SSinit still works for simple AR/MA orders.

python/statsforecast/arima.py
```diff
@@ -26,7 +26,19 @@
...
+    _VALID_DISTRIBUTIONS,
+    _quantiles,
+    error_params_from_model,
+    extract_dist_params,
+    guard_dist_tail,
+)
+
+# Finite stand-in for an undefined objective. An inf or nan makes scipy's line
+# search derive a NaN trial step and abort the fit, so keep the objective bounded.
+_UNDEFINED_OBJ = np.finfo(np.float64).max
```
nasaul: `_UNDEFINED_OBJ = np.finfo(np.float64).max` is still large enough to overflow to `inf`/`-inf` once it's divided or multiplied downstream, which reintroduces the non-finite propagation this change is meant to prevent.

Reproduced via this PR's own test suite: `RuntimeWarning: overflow encountered in divide` inside scipy's `_numdiff.py` for `test_cross_method_distribution_keys[AutoARIMA-t]` / `[AutoARIMA-ged]`, and `RuntimeWarning: overflow encountered in scalar multiply` at the AIC line (`value = 2*n_used*res.fun + ...`) during `test_forward_arima_models`.

A smaller bounded sentinel (e.g. `1e100` or similar, well below where scipy's finite-difference step or the AIC scaling can push it to `inf`) would close this gap.

python/statsforecast/arima.py
```diff
@@ -304,20 +328,22 @@ def armafn_laplace(p, x, trans, coef, mask, arma, mod, ncxreg, xreg, narma):
...
         sum_abs = np.nansum(np.abs(std_resid))
         if sum_abs <= 0:
-            return math.nan
+            return _UNDEFINED_OBJ
         b_hat = sum_abs / res[2]
         return math.log(b_hat) + 0.5 * res[1] / res[2]
 
     def armafn_t(p_ext, x, trans, coef, mask, arma, mod, ncxreg, xreg, narma):
         # p_ext = [arma_free..., log_sigma2, log_nu_m2]
+        if not np.isfinite(p_ext).all():
```
nasaul: Minor: this finiteness guard (`if not np.isfinite(p_ext).all(): return _UNDEFINED_OBJ`, and the matching `if not math.isfinite(obj): return _UNDEFINED_OBJ` at the end of each function) is duplicated verbatim across `armafn_t`, `armafn_skewnorm`, and `armafn_ged`. Worth factoring into a small shared helper/decorator so a future change to the guard logic doesn't need to be applied in three places and risk drifting out of sync.

nasaul (commented):
Automated review: verified against the PR branch by running the full test suite plus targeted repros. Most of the diff (arCheck/maInvert index fixes, seasonal-AR-check fix, ndiffs/myarima off-by-one fixes, CES/ETS/Theta box-bounds wiring) are legitimate bug fixes. Two regressions and one cleanup item found, left inline.

nasaul (approved):
LGTM

PR #1213, opened 2026-08-20
fix: avoid excessive ARIMA covariance workspace

### LINKED ISSUE #1034: [ARIMA] C++ max_size() error when using long `p` or `seasonalP` terms

With a time series with half-hourly granularity (48 samples per day), I am trying to run an ARIMA model but am running into limitations on the order of usable terms.

with (among other variations that raised the same error)

```python
ARIMA(
    order=(0, 0, 0),
    season_length=48,
    seasonal_order=(1, 1, 7),
)
```

When trying to fit, I get the error

```
❱   52 │   _arima.getQ0(phi, theta, res) 
...
ValueError: cannot create std::vector larger than max_size()
```

I tried to dig into the `arima.cpp` code to understand exactly where it's happening but that's outside of my wheelhouse. `max_size()` is generally a huge vector though so not sure what may be happening here.

I understand a half-hourly granularity is probably far beyond what ARIMA was meant to do, but I was surprised to run into such specific limits from the underlying code.

### Versions / Dependencies

Dependencies:

- python==3.10
- statsforecast==1.7.8
</details>

### Reproducible example

```python
# paste your code here
```

### Issue Severity

None

kiwoongyoon:
Title: fix: avoid excessive ARIMA covariance workspace

## Summary

- Compute the ARIMA initial state covariance with SciPy's discrete Lyapunov solver.
- Avoid the native temporary vector whose size grows roughly as the fourth power of the expanded ARMA order.
- Add a regression for the reported `(0, 0, 0)(1, 1, 7)[48]` fit and prediction.

Fixes #1034

## Test evidence

- The existing `getQ0` expected-value regression passes against the replacement.
- A focused 337-state assertion produces a finite covariance satisfying the discrete Lyapunov equation.
- Python compilation, Ruff on the changed production file, and `git diff --check` pass.
- The added fit-level pytest test awaits CI because local dependency setup could not obtain the CPython 3.12 `coreforecast` wheel in the network-restricted sandbox.

## Risks or notes for maintainers

High-order ARIMA optimization remains computationally expensive, but covariance initialization no longer requires the failing O(r⁴) workspace. SciPy is already a required dependency.

### DIFF
--- a/python/statsforecast/arima.py
+++ b/python/statsforecast/arima.py
@@ -22,6 +22,7 @@ import numpy as np
 import pandas as pd
 import statsmodels.api as sm
 from coreforecast.scalers import boxcox, boxcox_lambda, inv_boxcox
+from scipy.linalg import solve_discrete_lyapunov
 from scipy.optimize import minimize
 from scipy.signal import convolve
 from scipy.stats import norm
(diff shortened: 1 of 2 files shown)


kiwoongyoon:
Rebuilt this PR on the current upstream base in `45ab99ac6b` and reran the configured validation.

PR #1232, opened 2026-09-08
perf: reuse the ETS and Theta optimizer scratch buffers

jmoralez:
Both the ETS and Theta objective functions allocate and zero their working buffers on every Nelder–Mead evaluation, and `theta::forecast` additionally reallocates a growing state matrix on every timestep. This hoists all of it into caller-owned scratch, using the buffer/wrapper idiom `ces.cpp` already established, so the buffers are allocated once per fit rather than once per evaluation.

Why: `ets::ObjectiveFunction` and the `ets::Calc` it calls make seven allocations per evaluation, and `AutoETS` on a 200-point series with `season_length=24` runs ~5,500 evaluations per forecast; `theta::forecast`'s per-timestep copy is quadratic in the series length. Removing them cuts allocator traffic for an `AutoETS` + `AutoTheta` pair by ~78%, which is malloc-lock contention that users hit with `n_jobs > 1`, and cuts the Theta wall clock by 15–33%.

## What changes

- **`ets::CalcBuf`** takes `s`, `old_s`, `denom` and `f` from the caller; `Calc` stays as the allocating wrapper for the public API. Only `denom` is re-zeroed per call — every other buffer is fully written before it is read.
- **`Optimize`/`OptimizeDist` own a `Scratch`** that the objectives receive through `nm::NelderMead`'s existing variadic tail, so `nelder_mead.h` is untouched. Every `Scratch` is function-local and never `static`, so concurrent fits stay independent.
- **`theta::forecast` is allocation-free.** `update` splits into `update_step`, a pure transition over the previous row, and the horizon runs on two rolling 5-element rows seeded from `states.row(i - 1)`. O(h) instead of O(i + h).
- **`theta::calc_buf`** takes `denom` and `f` alongside the `states`/`e`/`amse` it already took, and `optimize`/`optimize_dist` own a `Scratch` the same way. `pegels_resid` keeps its own matrices — it hands them back to Python.
- **The objectives take `params` and `y` by `Eigen::Ref`** (thanks @nasaul). The optimizers forward a `Ref` but the objectives took a `const VectorXd&`, so a user-defined conversion copied the whole series on every evaluation — exactly the allocation this PR sets out to remove.
- **The distribution objectives zero `e`** rather than relying on `CalcBuf`/`calc_buf` having exactly one early exit that returns the `NA` sentinel (also @nasaul).

No arithmetic changes — the refactor is bit-exact by construction and verified as such. The bound signatures are unchanged: `_ets.calc`, `_ets.forecast`, `_ets.update`, `_theta.calc`, `_theta.forecast`, `_theta.update` and `_theta.pegels_resid` all keep their current signatures; the scratch-taking variants are additions.

One subtlety worth flagging for review: dropping the zero-fill of `state` makes the multiplicative-seasonality guard load-bearing, so it narrows from `state.tail(state.size() - start)` to `state.segment(start, p - start)`. That is equivalent today — the tail it also scanned was all zeros — but with a reused buffer a stale negative anywhere in the tail would have returned `+inf` and silently steered the optimizer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/ets.cpp
+++ b/src/ets.cpp
@@ -233,11 +233,27 @@ double Calc(Ref x, Ref e, Ref a_mse, int n_mse, CRef y, Component error,
                             beta, gamma, phi, m, s, old_s, denom, f);
 }
 
-double ObjectiveFunction(const VectorXd &params, const VectorXd &y, int n_state,
-                         Component error, Component trend, Component season,
-                         Criterion opt_crit, int n_mse, int m, bool opt_alpha,
-                         bool opt_beta, bool opt_gamma, bool opt_phi,
-                         double alpha, double beta, double gamma, double phi) {
+// Buffers shared by the objective evaluations of a single Optimize() call.
+// Function-local, never static: fits may run concurrently.
+struct Scratch {
+  VectorXd state;
+  VectorXd e;
+  VectorXd a_mse;
+  std::vector<double> s, old_s, denom, f;
+
+  Scratch(Eigen::Index n, Eigen::Index p, int m)
+      : state(p * (n + 1)), e(n),
+        // CalcBuf only refills a_mse's first n_mse entries
+        a_mse(VectorXd::Zero(30)), s(std::max(m, 24)), old_s(std::max(m, 24)),
+        denom(30), f(30) {}
+};
+
+double ObjectiveFunction(const VectorXd &params, Scratch &ws, const VectorXd &y,
+                         int n_state, Component error, Component trend,
+                         Component season, Criterion opt_crit, int n_mse, int m,
+                         bool opt_alpha, bool opt_beta, bool opt_gamma,
+                         bool opt_phi, double alpha, double beta, double gamma,
+                         double phi) {
   int j = 0;
   if (opt_alpha) {
     alpha = params(j++);
@@ -252,9 +268,8 @@ double ObjectiveFunction(const VectorXd &params, const VectorXd &y, int n_state,
     phi = params(j++);
   }
   auto n_params = params.size();
-  auto n = y.size();
   int p = n_state + (season != Component::Nothing);
-  VectorXd state = VectorXd::Zero(p * (n + 1));
+  VectorXd &state = ws.state;
   std::copy(params.data() + n_params - n_state, params.data() + n_params,
             state.data());
   if (season != Component::Nothing) {
@@ -264,15 +279,18 @@ double ObjectiveFunction(const VectorXd &params, const VectorXd &y, int n_state,
     state(n_state) =
         static_cast<double>(m * (season == Component::Multiplicative)) - sum;
     if (season == Component::Multiplicative &&
-        state.tail(state.size() - start).minCoeff() < 0.0) {
+        state.segment(start, p - start).minCoeff() < 0.0) {
       return std::numeric_limits<double>::infinity();
     }
   }
-  VectorXd a_mse = VectorXd::Zero(30);
-  VectorXd e = VectorXd::Zero(n);
-  double lik = Calc<VectorXd &, const VectorXd &>(state, e, a_mse, n_mse, y,
-                                                  error, trend, season, alpha,
-                                                  beta, gamma, phi, m);
+  VectorXd &a_mse = ws.a_mse;
+  VectorXd &e = ws.e;
+  // CalcBuf can return early, leaving e's tail untouched; Sigma and MAE read
+  // all of it.
+  e.setZero();
+  double lik = CalcBuf<VectorXd &, const VectorXd &>(
+      state, e, a_mse, n_mse, y, error, trend, season, alpha, beta, gamma, phi,
+      m, ws.s, ws.old_s, ws.denom, ws.f);
   lik = std::max(lik, -1e10);
   if (std::isnan(lik) || std::abs(lik + 99999.0) < 1e-7) {
     lik = -std::numeric_limits<double>::infinity();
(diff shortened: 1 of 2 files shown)


nasaul:
Low severity, but it affects the headline claim: the per-evaluation `y` copy this PR sets out to remove is still there.

`Optimize`/`OptimizeDist` (`src/ets.cpp:302`) and `optimize`/`optimize_dist` (`src/theta.cpp:210`, `src/theta.cpp:260`) take `const Eigen::Ref<const VectorXd> &y`, but `ObjectiveFunction`/`target_fn` take `const VectorXd &y`. `NelderMead` forwards the `Ref` into a `const VectorXd&` parameter, which requires a user-defined conversion — so a fresh heap-allocated `VectorXd` copy of the whole series is materialised on *every* objective evaluation. For the `AutoETS` example in the description that is roughly 5,500 allocations of 200 doubles per forecast, so the "allocator traffic drops from ~59,000 calls to a handful" claim does not hold yet.

Fix is one line per objective: change the `y` parameter to `const Eigen::Ref<const VectorXd>&` as well. (Same applies to `const VectorXd &params` receiving `simplex.row(i)`, though that copy is small.)

nasaul:
Second low-severity note: the "`e` is fully written past the guard" invariant is now load-bearing but unenforced.

`ObjectiveFunctionDist` (`src/ets.cpp:370`) and `target_fn_dist` (`src/theta.cpp:222`) deliberately skip zeroing `ws.e`, relying on `CalcBuf`/`calc_buf` having exactly one early exit and that exit returning the sentinel `NA = -99999.0`, which the `abs(lik + 99999.0) < 1e-7` guard catches. That holds today — I traced it.

But it is a coupling between two functions held together only by a comment. Any future early return from `CalcBuf` that does not return `NA` (say a new parameter-validity bail-out returning `HUGE_N` or `inf`) would silently feed the previous evaluation's residuals into `negloglik_*`, producing a plausible-looking but wrong objective value with no crash and no NaN.

Cheap hardening: return `NA` from every exit, or assert it.

jmoralez:
> [the per-evaluation `y` copy this PR sets out to remove is still there](https://github.com/Nixtla/statsforecast/pull/1232#issuecomment-5594063170)

You're right, and thanks — fixed in 28bd837. I'd spotted this while writing the PR and left it out because it was worth only ~0.4% wall clock, which was the wrong test to apply: the claim it undercuts is about allocator *count*, not time.

Changed `y` and `params` to `const Eigen::Ref<const VectorXd>&` in all four objectives. `params` turns out to matter as much as `y` by count — one copy per evaluation either way, just fewer bytes — and it's free to fix because `simplex` is row-major, so `simplex.row(i)` is contiguous and `Ref<const VectorXd>` binds as a view rather than falling back to a copy.

Measured over one `AutoETS` + one `AutoTheta` forecast, differencing an N-iteration run against N=0:

| | allocator calls per pair |
|---|---|
| main | 60,754 |
| before your comment | 18,894 |
| after | 13,299 |

So "drops to a handful" was wrong even after the fix, and I've corrected the description. DHAT puts ~80% of the remaining ~13,300 in four sites inside `nm::NelderMead` itself — the per-iteration `x_o`/`x_r`/`x_e`/`x_oc`/`x_ic` temporaries and the `Clamp` results — which are in `nelder_mead.h` and out of scope here.

Chasing this also turned up a bigger correction that I've made to the description: the ~5% `AutoETS`/`HoltWinters` gain I originally quoted was measured on the #1195 branch, whose CMake build enables LTO. On `main`'s non-LTO build the ETS half is wall-clock-neutral (−0.9%), because the `VectorXd::Zero` fill compiles to a glibc `memset` that is nearly free at this size rather than to inlined stores. The Theta gains are real under both backends. CodSpeed's `×2.2` is the same cost-model artifact as on #1195, running in the other direction.

`ces_target_fn` has the identical problem and additionally copies `init_states` per evaluation — I've noted it for the follow-up PR.

jmoralez:
> [the "`e` is fully written past the guard" invariant is now load-bearing but unenforced](https://github.com/Nixtla/statsforecast/pull/1232#issuecomment-5594071224)

Agreed, and your trace is right — one early exit, returning `NA`, caught by the `abs(lik + 99999.0) < 1e-7` guard. Addressed in fb3c9b7, though not with either of the two options you suggested.

"Return `NA` from every exit" is vacuous today (there is only the one), and an assert would only fire in a debug build, so neither actually constrains the future early return you're worried about. Instead I dropped the dependency: `ObjectiveFunctionDist` and `target_fn_dist` now zero `e` per evaluation, exactly as the non-distribution `ObjectiveFunction` already did. No caller reasoning about `CalcBuf`'s exits, and the asymmetry between the two objectives — which was itself a trap — is gone.

The reason I'd avoided this originally was cost, and that turned out not to apply: `optimize_dist` runs only for `distribution != "normal"`, so none of the benchmarked paths touch it. The `n`-double fill is free where it lands.

I did consider the third option, zeroing `e`'s tail inside `CalcBuf` on the `NA` path so the invariant holds by construction. Rejected it: `_ets.calc` is bound to Python and `pegelsresid_C` passes a NaN-filled `e`, so a truncated fit's residuals would change from NaN to 0 — user-visible, and the bit-exactness check catches it.

nasaul (approved):
Traced every buffer whose zero-init was dropped; no path reads stale or uninitialized data.

- `src/ets.cpp` — `CalcBuf` re-zeroes `denom`/`a_mse` over `[0, n_mse)` after the `min(n_mse, 30)` clamp, matching exactly what the accumulation loop touches. `s`/`old_s` are only indexed in `[0, m)` under `season != Nothing`, the same branch that seeds them. `f` is fully written by `Forecast(h = n_mse)` before any read. `ws.state[0..p)` is rewritten from `params` each evaluation.
- The `state.tail(...)` → `state.segment(start, p - start)` narrowing is equivalent: the dropped tail was zero-filled, and a `minCoeff` of `0.0` never trips the `< 0.0` test.
- `src/theta.cpp` — `update_step` reads `prev[0..3]` into locals before writing `cur`, and `cur[2]` reads the already-written `cur[1]`/`cur[3]`, preserving the original ordering. `prev`/`cur` don't alias. `target_fn` genuinely doesn't need `e.setZero()`.
- `nm::NelderMead` forwards `Scratch` as an lvalue reference (no copy), and each `Scratch` is function-local, so concurrent fits stay independent.
- `nmse` is validated to `1..30` in Python, so the degenerate `n_mse <= 0` case that would read an unwritten `f[0]` is unreachable.

One nit, not blocking: `theta::forecast` hardcodes 5 columns (`std::array<double, 5>`) where it previously used `states.cols()`, while the pybind signature still accepts an arbitrary-width matrix. `cols > 5` is identical (those columns were never read) and `cols < 5` was already broken pre-PR — documentation debt rather than a defect.

### Verification

- Both files compile clean under `clang++ -std=c++20 -fsyntax-only`.
- Built main and the PR head at `-O2` and diffed 59 forecast/fitted vectors across 5 series shapes, `season_length ∈ {1, 12, 24}`, `AutoETS`/`AutoTheta`/`DynamicOptimizedTheta`, plus the laplace/t/skew-normal/ged paths: 0 differences, bit-exact.
- Stronger check on the dropped zero-inits: patched a `poison()` that NaN-fills every scratch buffer at the top of every objective evaluation, rebuilt, re-ran — still 0 differences, and `test_ets`/`test_theta`/`test_distributions`/`test_models` → 375 passed. Any stale read would have propagated NaN.
- Unpoisoned PR build: 169 passed.

